PR #401, opened 2019-09-19
Report "CAM error" when SensorFailure occurs

bmerry:
These errors are invariably due to problems in katportal/katstore, and
this will hopefully direct operators to the right people for debugging.

### DIFF
--- a/katsdpcontroller/sdpcontroller.py
+++ b/katsdpcontroller/sdpcontroller.py
@@ -1436,7 +1436,7 @@ class SDPControllerServer(DeviceServer):
             config_dict = await product_config.update_from_sensors(config_dict)
             config_dict = product_config.normalise(config_dict)
         except product_config.SensorFailure as error:
-            retmsg = "CAM error: {}".format(error)
+            retmsg = "Erorr retrieving sensor data from CAM: {}".format(error)
             product_logger.error(retmsg)
             raise FailReply(retmsg) from error
         except (ValueError, jsonschema.ValidationError) as error:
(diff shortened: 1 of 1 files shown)


katsdpcontroller/sdpcontroller.py
```diff
@@ -1435,10 +1435,14 @@ def remove_product(product):
             product_config.validate(config_dict)
             config_dict = await product_config.update_from_sensors(config_dict)
             config_dict = product_config.normalise(config_dict)
-        except (ValueError, product_config.SensorFailure, jsonschema.ValidationError) as error:
+        except product_config.SensorFailure as error:
+            retmsg = "CAM error: {}".format(error)
```
sratcliffe: I know I asked for this, but are we being too blunt in calling this a CAM error :) Apart from katstore being completely unavailable or timeouts on sensor requests, what else could trigger SensorFailure ?

bmerry: Easiest answer to that is to search for it in product_config.py. I guess some of the cases (like the sensor not existing) could be due to CBF as well.

sratcliffe: OK - how about "Sensor error (possible CAM problem): {}" ?

martinslabber: Maybe be a bit more helpful and say: "Could not get sensor DATA from CAM:", DATA is not clear but sure you can substitute.

sratcliffe: Final bikeshed attempt: "Error retrieving sensor data from CAM: {}"

bmerry: Done

PR #402, opened 2019-09-20
Lower bound to timeplot CPU allocation.

sratcliffe:
This was basically determined by inspection of a various running scenarios, particularly 1k and 4 antenna modes.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -616,7 +616,9 @@ def _make_timeplot_correlator(g, config, spectral_name):
 
     # Exact requirement not known (also depends on number of users). Give it
     # 2 CPUs (max it can use) for 16 antennas, 32K channels and scale from there.
-    cpus = 2 * min(1.0, spectral_info.n_vis / _N16_32)
+    # Lower bound (from inspection) to cater for fixed overheads.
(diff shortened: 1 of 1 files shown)


katsdpcontroller/generator.py
```diff
@@ -616,7 +616,9 @@ def _make_timeplot_correlator(g, config, spectral_name):
 
     # Exact requirement not known (also depends on number of users). Give it
     # 2 CPUs (max it can use) for 16 antennas, 32K channels and scale from there.
-    cpus = 2 * min(1.0, spectral_info.n_vis / _N16_32)
+    # Lower bound (from inspection) to cater for fixed overheads.
+    cpus = max(2 * min(1.0, spectral_info.n_vis / _N16_32), 0.3)
```
bmerry: Have you looked into the effect of integration time? We originally thought we'd always be doing 2s integrations, but now it varies between 1s (holography) and 8s and it might make sense to combine that into the scaling.

Nevertheless a minimum is probably wise.

sratcliffe: Good point - let's roll out this minimum and then see if we still have scaling mismatches (up or down) over a couple of week period.

PR #403, opened 2019-09-26
Take max_channels into account in computing writer chunking

bmerry:
This will give larger chunks for the spectral imager data stream,
hopefully improving performance (closes SPR1-227). The max memory usage
for the writers is increased to 32GB. In practice I think that's only
going to affect large 32K-channel arrays when writing visibilities for
spectral imaging, because in other cases either the time length or the
chunk size limit kicks in first.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -43,6 +43,14 @@ IMAGES = frozenset([
     'katsdpmetawriter',
     'katsdptelstate'
 ])
+#: Number of bytes per complex visibility
+BYTES_PER_VIS = 8
+#: Number of bytes per per-visibility flag mask
+BYTES_PER_FLAG = 1
+#: Number of bytes per per-visibility weight
+BYTES_PER_WEIGHT = 1
+#: Number of bytes per vis-flags-weights combination
+BYTES_PER_VFW = BYTES_PER_VIS + BYTES_PER_FLAG + BYTES_PER_WEIGHT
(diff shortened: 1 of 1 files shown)


katsdpcontroller/generator.py
```diff
@@ -1122,7 +1131,7 @@ def _make_vis_writer(g, config, name, s3_name, local, prefix=None, max_channels=
     vis_writer.cpus = min(3, 2 * info.n_vis / _N32_32)
 
     workers = 4
-    max_accum_dumps = _writer_max_accum_dumps(config, name)
+    max_accum_dumps = _writer_max_accum_dumps(config, name, 10, max_channels)
```
mauch: Perhaps add a comment (or use a constant) to indicate the meaning of `bytes_per_vis` here.

katsdpcontroller/generator.py
```diff
@@ -1189,7 +1198,7 @@ def _make_flag_writer(g, config, name, l0_name, s3_name, local, prefix=None, max
     flags_src = find_node(g, 'multicast.' + name)
     n_substreams = flags_src.n_addresses
     workers = 4
-    max_accum_dumps = _writer_max_accum_dumps(config, l0_name)
+    max_accum_dumps = _writer_max_accum_dumps(config, l0_name, 1, max_channels)
```
mauch: And here.

katsdpcontroller/generator.py
bmerry: Done.

PR #405, opened 2019-09-26
Improve launch speed for batch tasks

bmerry:
Previously each time a group was launched, all offers were returned
(either declined by launching zero tasks, or used by launching a
non-zero number of tasks). Now offers from unused agents are retained if
there are more launch groups in the queue (which is typical for spectral
imaging).

This partly addresses SPR1-231. It may still be possible to improve
further by using an offer filter for the agents where launching does
happen, so that the remaining resources are immediately re-offered.


bmerry:
This is theoretical so far, but I'll give it a test tomorrow.

PR #414, opened 2019-10-16
Make 'Lost connection' display on dashboard work again

bmerry:
It was being thwarted by the proxying: it redirected to the master
controller's web server, and even though that returned a 405, dash would
treat this as completion and reset the loading state.

Hack around that by delaying the response by 2s, so that dash always
believes it has stale data.

Closes SPR1-257.

### DIFF
--- a/katsdpcontroller/web.py
+++ b/katsdpcontroller/web.py
@@ -40,7 +40,7 @@ async def _prometheus_handler(request: web.Request) -> web.Response:
     response = await prometheus_async.aio.web.server_stats(request)
     if response.status == 200:
         # Avoid spamming logs (feeds into web_utils.AccessLogger).
-        response.log_level = logging.DEBUG
+        response['log_level'] = logging.DEBUG
     return response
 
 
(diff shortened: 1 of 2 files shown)


katsdpcontroller/web.py
```diff
@@ -103,6 +103,20 @@ def _rotate_handler(request: web.Request) -> dict:
...
+    """Make dashboard wait for a response.
+
+    This seems to be the easiest way to trick it into thinking it's not
+    connected. We just need to put it to sleep for longer than the
+    refresh interval.
+    """
+    await asyncio.sleep(2)
+    response = web.json_response({}, status=404)
+    # Avoid spamming the logs many times a second
+    response['log_level'] = logging.DEBUG
```
sratcliffe: Confused :) - how does setting the log level to DEBUG help the log spamming ?

bmerry: It's the log level at which the details of the request are logged. And since only messages at INFO or higher actually end up in Logtrail etc, it will be invisible.

sratcliffe: So works in opposition to what you might expect when setting the log_level on an object

PR #418, opened 2019-10-21
Limit imaging to observations with enough time on target

bmerry:
This required some restructuring of the "target cache". Previously it
cached full target list, but spectral and continuum no longer image the
same set of targets. It's been replaced by a `TargetMapper` that is
responsible just for assigning a name to a target in an idempotent way.
This actually simplifies the interfaces since `_get_targets` (formerly
known as `_unique_targets`) no longer has any responsibility for naming.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -75,7 +75,7 @@ SPECTRAL_OBJECT_CHANNELS = 64
 #: Minimum observation time for continuum imager (seconds)
 DEFAULT_CONTINUUM_MIN_TIME = 15 * 60     # 15 minutes
 #: Minimum observation time for spectral imager (seconds)
-DEFAULT_SPECTRAL_MIN_TIME = 2 * 3600     # hours
+DEFAULT_SPECTRAL_MIN_TIME = 2 * 3600     # 2 hours
 
 logger = logging.getLogger(__name__)
 
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -1770,7 +1770,7 @@ def _make_continuum_imager(g, config, capture_block_id, name, telstate, target_m
             '--telstate', '{endpoints[telstate_telstate]}',
             '--access-key', '{resolver.s3_config[continuum][read][access_key]}',
             '--secret-key', '{resolver.s3_config[continuum][read][secret_key]}',
-            '--select', f'scans="track"; corrprods="cross"; targets=[{target.description!r}]'
+            '--select', f'scans="track"; corrprods="cross"; targets=[{target.description!r}]',
             '--capture-block-id', capture_block_id,
             '--output-id', name,
             '--telstate-id', telstate.join(name, target_name),
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -72,6 +72,10 @@ CONFIG_VOL = scheduler.VolumeRequest('config', '/var/kat/config', 'RW')
 WRITER_OBJECT_SIZE = 20e6    # 20 MB
 #: Maximum channels per chunk for spectral imager
 SPECTRAL_OBJECT_CHANNELS = 64
+#: Minimum observation time for continuum imager (seconds)
+DEFAULT_CONTINUUM_MIN_TIME = 15 * 60     # 15 minutes
+#: Minimum observation time for spectral imager (seconds)
+DEFAULT_SPECTRAL_MIN_TIME = 2 * 3600     # hours
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


bmerry:
Closes SPR1-266.

katsdpcontroller/generator.py
```diff
@@ -1642,30 +1644,45 @@ def _sky_model_url(data_url, continuum_name, target):
...
+        if target in self._cache:
+            return self._cache[target]
+        name = re.sub(r'[^-A-Za-z0-9_]', '_', target.name)
+        if name in self._used:
+            i = 1
+            while '{}_{}'.format(name, i) in self._used:
+                i += 1
+            name = '{}_{}'.format(name, i)
+        self._cache[target] = name
+        self._used.add(name)
```
sratcliffe: Not that it matters much since, as you say, this is not thread safe anyway, but wouldn't reserving the name before adding it to the cache make more sense.

katsdpcontroller/generator.py
```diff
@@ -1718,29 +1737,31 @@ def _render_continuum_parameters(parameters):
...
     cpus = _continuum_imager_cpus(config)
-    targets = _unique_targets(config, capture_block_id, name, telstate, target_cache)
+    # Image targets observed for at least 15 minutes
+    targets = _get_targets(config, capture_block_id, name, telstate, 15 * 60)
 
     # katdal doesn't support selection by target description, and selection by
     # name can be ambiguous (if there are multiple targets with the same
     # name), so we select by index.
-    catalogue = _get_targets(telstate, capture_block_id, l0_stream)
+    catalogue = _get_data_set(telstate, capture_block_id, l0_stream).catalogue
```
sratcliffe: Do we really need to do this again ? _get_targets has just trawled the catalogue and picked targets from it - how would they disappear between these calls ?

katsdpcontroller/generator.py
```diff
@@ -1718,29 +1737,31 @@ def _render_continuum_parameters(parameters):
...
     output = config['outputs'][name]
     l1_flags_name = output['src_streams'][0]
     l0_name = config['outputs'][l1_flags_name]['src_streams'][0]
     l0_info = L0Info(config, l0_name)
     l0_stream = name + '.' + l0_name
     data_url = _stream_url(capture_block_id, l0_stream)
     cpus = _continuum_imager_cpus(config)
-    targets = _unique_targets(config, capture_block_id, name, telstate, target_cache)
+    # Image targets observed for at least 15 minutes
+    targets = _get_targets(config, capture_block_id, name, telstate, 15 * 60)
```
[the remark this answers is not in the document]
bmerry: Fair enough, will do.

I'm also thinking it might be nice (mainly for development) to let them be overridden from config - mostly for development where I don't want to have to do a 2 hour simulation just to check the spectral imager plumbing.

katsdpcontroller/generator.py
bmerry: I don't follow - what change do you want to make?

katsdpcontroller/generator.py
bmerry: Yeah, I wasn't totally happy with this either. I think I might just fix https://skaafrica.atlassian.net/browse/SR-1819, after which it won't be necessary to get the catalogue again to look up indices.

katsdpcontroller/generator.py
sratcliffe: Nothing important, just swapping the order of self._cache... and self._used...

katsdpcontroller/generator.py
sratcliffe: Or we could return the indices together with the target mapping from the _get_data_set call

bmerry: That was my fallback idea too.

bmerry:
I've implemented configurability.

katsdpcontroller/generator.py
```diff
@@ -72,6 +72,10 @@
 WRITER_OBJECT_SIZE = 20e6    # 20 MB
 #: Maximum channels per chunk for spectral imager
 SPECTRAL_OBJECT_CHANNELS = 64
+#: Minimum observation time for continuum imager (seconds)
+DEFAULT_CONTINUUM_MIN_TIME = 15 * 60     # 15 minutes
+#: Minimum observation time for spectral imager (seconds)
+DEFAULT_SPECTRAL_MIN_TIME = 2 * 3600     # hours
```
sratcliffe: # 2 hours ?

katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -42,6 +42,10 @@
             "type": "integer",
             "minimum": 1
         },
+        "nonneg_number": {
+            "type": "number",
+            "minimum": 0.0
```
sratcliffe: I wonder if we should actually allow 0 - could we get a target in the list that is never actually observed ? Some sort of sensible minimum like 60second could make sense, although I am sure we could imagine a pathological case in which this wouldn't work.

katsdpcontroller/generator.py
bmerry: Yep. Fixed.

katsdpcontroller/schemas/product_config.json.j2
bmerry: Good point - strictly positive makes more sense. Not sure why I thought it would be a good idea to allow 0, as it'll probably cause things to crash. I've changed it, although I've kept the nonneg_number definition since I put it to use elsewhere.

PR #421, opened 2019-10-21
Stop making titanx special for normalise_gpu_name

bmerry:
See SPR1-248. Requires ska-sa/katsdpingest#330 to be merged first.


bmerry:
I added another commit for update-local-registry.

PR #427, opened 2019-11-20
Expose subarray `state` sensor to Prometheus

bmerry:
This will allow Grafana dashboards to show the subarray product state
(via a value-to-text mapping). Since Prometheus doesn't support string
values, enum/discrete sensors are mapped to an integer according to the
position in the list of legal values.

### DIFF
--- a/katsdpcontroller/sensor_proxy.py
+++ b/katsdpcontroller/sensor_proxy.py
@@ -2,7 +2,7 @@
 
 import logging
 import enum
-from typing import Tuple, Dict, Callable, Type, Mapping, Iterable, Sequence, Union, Optional
+from typing import List, Tuple, Dict, Callable, Type, Mapping, Iterable, Sequence, Union, Optional
 
 import aiokatcp
 import prometheus_client       # noqa: F401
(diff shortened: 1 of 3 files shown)


katsdpcontroller/sensor_proxy.py
```diff
@@ -65,7 +78,7 @@ def __init__(self, sensor: aiokatcp.Sensor,
 
     def __call__(self, sensor: aiokatcp.Sensor, reading: aiokatcp.Reading) -> None:
         valid = reading.status.valid_value()
-        value = float(reading.value) if valid else 0.0
+        value = _reading_to_float(reading) if valid else 0.0
```
sratcliffe: Just checking the interplay between !valid and the Enum index failing since they map to 0.0 and -1.0 respectively. Why have two states for invalid effectively ?

bmerry: Good question. They're slightly different things: !valid means I couldn't find out the value (e.g. it hasn't had an initial value set, or I've lost my katcp connection), enum index failing indicates a bug in the sensor provider. I think the right solution in the former case is to stop listing the metric to Prometheus entirely (Prometheus 1.x didn't handle that well, but I think Prometheus 2 is better). So far we've been using the auxiliary `xxx_status` metric, and in theory we could be using that to filter the actual metric in the query, but in practice that makes all queries far too complex to maintain.

sratcliffe: Thanks for the explanation

bmerry:
I've opened https://skaafrica.atlassian.net/browse/SPR1-315 to deal with better handling of "invalid" sensors.

PR #431, opened 2019-11-26
Build haproxy from source

bmerry:
The PPA I was using released a new version and **deleted** the version
we've pinned. So rather than have the carpet removed from under us
every release, I'm building haproxy from a source download. This might
even make the image slightly smaller since only the binary is copied to
the final image, not the docs.


bmerry:
@ludwigschwardt would you mind reviewing while @sratcliffe is away?

PR #437, opened 2019-12-10
Add separate 'postprocessing' state

bmerry:
Subarray products now go into 'deconfiguring' when ?product-deconfigure
is called, and into 'postprocessing' when it returns (indicating the
realtime processes including cam2telstate have been killed). From CAM's
point of view, 'postprocessing' subarray products are no longer CAM's
problem and have no effect other than on resources.

This should help with detecting things like
(a) CAM configuring a new subarray before properly deconfiguring the
old one.
(b) SDP gets jammed up in the 'deconfiguring' state, versus just ticking
along in 'postprocessing' state.

In future we could also use it to do things like refuse to build a new
subarray product with the same name wildcard as an existing one (other
than those in 'postprocessing') and CAM could use it to do
reconciliation with its internal state.

I've bumped the device server minor versions, but this will still need
an ECP to update the ICD.

### DIFF
--- a/katsdpcontroller/controller.py
+++ b/katsdpcontroller/controller.py
@@ -130,7 +130,8 @@ class ProductState(scheduler.OrderedEnum):
     - CONFIGURING -> IDLE (via product-configure)
     - IDLE -> CAPTURING (via capture-init)
     - CAPTURING -> IDLE (via capture-done)
-    - CONFIGURING/IDLE/CAPTURING/ERROR -> DECONFIGURING -> POSTPROCESSING -> DEAD (via product-deconfigure)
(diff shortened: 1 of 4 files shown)


bmerry:
@ajoubertza, I've assigned you just to review the interface change - I don't expect you to review the implementation.

@zjassiem for you information - this will need an ICD change eventually.

bmerry:
Note: this is a case where we probably should drain the system and restart the master controller. Having the product controller and master controller disagree about the ProductState enum *might* cause problems (I'm not sure).

Before deploying we should also update the Grafana Overview dashboard with the new state.

PR #439, opened 2019-12-12
Implement Mesos task reconciliation

bmerry:
This is a fairly simple-minded version, but hopefully that makes it
robust. See http://mesos.apache.org/documentation/latest/reconciliation/
for a description of reconciliation and Mesos' recommendations for more
sophisticated variants of it.

Closes SPR1-326 and SPR1-327.

### DIFF
--- a/katsdpcontroller/scheduler.py
+++ b/katsdpcontroller/scheduler.py
@@ -3318,49 +3318,6 @@ class Scheduler(pymesos.Scheduler):
         status = await self._loop.run_in_executor(None, lambda: self._driver.join())
         return status
 
-    async def get_master_and_slaves(self, timeout=None):
-        """Obtain a list of slave hostnames from the master.
-
-        Parameters
-        ----------
-        timeout : float, optional
-            Timeout for HTTP connection to the master
-
-        Raises
-        ------
-        aiohttp.client.ClientError
-            If there was an HTTP connection problem (including timeout)
-        ValueError
-            If the HTTP response from the master was malformed
-        ValueError
-            If there is no current master
-
-        Returns
-        -------
-        master : list of str
-            Hostname of the master
-        slaves : list of str
-            Hostnames of slaves
-        """
-        if self._driver is None:
-            raise ValueError('No driver is set')
-        # Need to copy, because the driver runs in a separate thread
-        # (self.master is a property that takes a lock).
-        master = self._driver.master
-        if master is None:
-            raise ValueError('No master is set')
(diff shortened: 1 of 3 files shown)


katsdpcontroller/scheduler.py
```diff
@@ -2640,6 +2647,18 @@ def statusUpdate(self, driver, status):
...
+                    and status.state != 'TASK_KILLING'
+                    and status.state not in TERMINAL_STATUSES
+                    and self._loop.time() > task.kill_sent_time + self.kill_timeout):
+                logger.warning('Retrying kill on %s (task ID %s)',
+                               task.name, status.task_id.value)
+                self._driver.killTask(status.task_id)
+        else:
+            if status.state not in TERMINAL_STATUSES:
+                logger.warning('Received status update for unknown task %s, killing it',
+                               status.task_id.value)
```
sratcliffe: Is the timestamp field in the status of any use ? i.e. does it tell us when the task was launched or only when this update was sent. Just wondering if there is any further debug info we can put into the warning to try and track down why we have an unknown task.

katsdpcontroller/scheduler.py
```diff
@@ -3278,6 +3297,34 @@ def depends_resources(cls, data):
...
+            try:
+                if explicit:
+                    tasks = [{'task_id': task_id} for task_id in self._active]
+                    logger.debug('Requesting explicit reconciliation of %d tasks', len(tasks))
+                else:
+                    tasks = []
+                    logger.debug('Requesting implicit reconciliation')
+                self._driver.reconcileTasks(tasks)
+            except Exception:
+                logger.warning('Exception during task reconciliation', exc_info=True)
```
sratcliffe: Why so general - what sort of failure modes to we expected from reconcileTasks ?

bmerry: I'm not expecting any failures, and I hope this will never run. But if one call to reconcileTasks fails, I want to keep trying rather than exiting the loop.

katsdpcontroller/scheduler.py
bmerry: I think the timestamp will just tell us when the update was sent (or maybe when the reconciliation request was received). There probably isn't much else we can get:

> Status updates that result from reconciliation requests will their reason field set to REASON_RECONCILIATION. Note that most of the other fields in the returned TaskStatus message will not be set: for example, reconciliation cannot be used to retrieve the labels or data fields associated with a running task.

PR #442, opened 2019-12-17
Disable spectral imager

bmerry:
This is a hotfix to prevent the spectral imager running until some
issues in the batch scheduler can be dealt with (SPR1-331 and SPR1-332
in particular).


bmerry:
This shouldn't be merged - just making a draft PR to get a quick sanity check on it.

bmerry:
The Jenkins build failed due to changes in katdal, so I've built and pushed to the registry manually.

PR #443, opened 2019-12-17
Fix explicit Mesos reconciliation

bmerry:
The request wasn't properly constructed according to the schema.

Relates to SPR1-326.


bmerry:
Somehow I'd missed this when I tested it the first time - the generic exception handler you were asking about indeed logged an error for this.

PR #447, opened 2020-01-09
Increase max run time for spectral imager to 6 hours

bmerry:
Closes SPR1-351.


bmerry:
@sratcliffe I've merged. We should deploy this before tonight but I guess we should wait to see what Tom finds in the continuum imager?

sratcliffe:
Definitely - I will check with Sean to get a gap early afternoon to do the deployment.

PR #449, opened 2020-01-10
Temporary fix to cater for long capture-done delays in vis_writers

sratcliffe:
I know this is not best practice, but at the moment we often see the subarray ending up in error state waiting for the spectral vis_writer to complete capture_done (typically takes 3 to 5 minutes). Although we may end up occasionally waiting longer than needed when it has properly wedged, I would prefer to make things a bit more robust while we test and understand 32k performance.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -37,7 +37,7 @@ CAPTURE_TRANSITIONS = {
         KatcpTransition('capture-init', '{capture_block_id}', timeout=30)
     ],
     CaptureBlockState.BURNDOWN: [
-        KatcpTransition('capture-done', timeout=120)
+        KatcpTransition('capture-done', timeout=600)
     ]
 }
 #: Docker images that may appear in the logical graph (used set to Docker image metadata)


bmerry:
I think this is okay as a temporary fix, but you'll also need CAM to increase their timeouts as I don't think they'll wait this long for us (400s rings a bell).

Do you think this is a throughput or a latency problem on Ceph? It may be that we can improve things just by using more threads in the vis writer (the default of 4 was based on the assumption of local storage - we should probably be using 10x that for S3), but on the other hand that may just thrash Ceph harder.

sratcliffe:
I really don't know what is causing this - Ceph seems fine at the moment. I need to run some performance testing on it once this observation is done to see how it behaves. At the moment the spectral vis_writer seems totally broken - continually dropping and reporting full ringbuffer - which is not something I have seen before.

ajoubertza:
> CAM to increase their timeouts as I don't think they'll wait this long for us (400s rings a bell).

Yes, 400 seconds is correct.

> complete capture_done (typically takes 3 to 5 minutes). Although we may end up occasionally waiting longer than needed when it has properly wedged

When it is "properly wedged" does it ever complete?  I'd prefer not to change the CAM timeouts, if 400 seconds is enough.  It requires changes in 3 repos...

sratcliffe:
@ajoubertza - I don't think we should change anything on the CAM side - this was a bit of a knee jerk on my side, and I think we have a good bunch of options to improve this delay and bring things back under control. Will let you know if we get to the point of not being able to improve, and thus require changing the delay.

bmerry:
@sratcliffe is this still an issue now that we've given the writers more threads? If so can we revert this?

sratcliffe:
Still checking through the logs a bit to make sure on this...

PR #453, opened 2020-01-13
Make spectral imager use tmp file on disk

bmerry:
This makes it about 10% slower (at the moment - may get more significant
with further optimisation), but keeps memory bounded. A quick test
suggests 9GB is probably enough, but allowing 15GB because we can afford
it and it gives some spare room that can be used for caching. It's still
more than I think ought to be getting used given the design, but that
may be because many channels are being loaded and processed at once.

Depends on ska-sa/katsdpdockerbase#44.


bmerry:
Don't merge yet - I still need to test that it actually works once the dockerbase change is merged and the image rebuilt.

PR #454, opened 2020-01-13
Support separate URLs for read and write in s3_config

bmerry:
For compatibility, each archive can be specified as before, or url's can
be specified separately in the `read` and `write` sections and not in
the top level (making for a somewhat complex JSON schema). There is a
schema fix, where previously the access key and secret key weren't
marked required even though they are.

Internally, the old format is rewritten into the new format by the
product controller when it starts up.

This is slightly brittle in terms of backwards/forwards compatibility:
once we update the config file, we'll only be able to use the new
version of the product controller. It'll also need a new
katsdpdatawriter to actually have the desired effect, although an old
one will still work but use the read endpoint for writing.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -437,7 +437,7 @@ def _make_cam2telstate(g, config, name):
 
 def _make_meta_writer(g, config):
     def make_meta_writer_config(task, resolver):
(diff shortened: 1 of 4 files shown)


bmerry:
@sratcliffe reminder that we'll need this before we can test split read/write.

PR #455, opened 2020-01-15
Increase spectral channels from 64 to 128

bmerry:
Testing (on an otherwise idle system) shows about 39m to process 64
channels and 64m to process 128 i.e. ~20% throughput increase. This is
because there are fixed costs such as opening the dataset and computing
parallactic angles and UVW coordinates.


sratcliffe:
Thanks - let's update this with real world numbers when we start running things in anger.

PR #456, opened 2020-01-15
Fix broken-ness from split-s3 change

bmerry:
It seems the unit tests weren't adequate to discover a number of issues. See individual commits. This is a blocker for deployment.

### DIFF
--- a/katsdpcontroller/schemas/s3_config.json
+++ b/katsdpcontroller/schemas/s3_config.json
@@ -1,65 +1,80 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
-    "definitions": {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$defs": {
         "key": {"type": "string"},
-        "key_pair": {
+        "has_keys": { "required": ["access_key", "secret_key"] },
+        "has_url": { "required": ["url"] },
+        "no_url": { "not": { "required": ["url"] } },
+        "section": {
+            "$comment": "A read or write section in a store (no required properties)",
             "type": "object",
             "additionalProperties": false,
             "properties": {
-                "access_key": {"$ref": "#/definitions/key"},
-                "secret_key": {"$ref": "#/definitions/key"}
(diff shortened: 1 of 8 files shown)


katsdpcontroller/schemas/s3_config.json
```diff
@@ -1,65 +1,80 @@
...
+            "if": { "$ref": "#/$defs/has_url" },
+            "then": {
+                "$comment": "If an URL is passed to store, URLs can't be used in sections",
+                "properties": {
+                    "read": {"$ref": "#/$defs/no_url"},
+                    "write": {"$ref": "#/$defs/no_url"}
+                }
+            },
+            "else": {
+                "$comment": "If not URL is passed to store, URLs must be used in sections",
```
sratcliffe: Perhaps not -> no rather

katsdpcontroller/schemas/s3_config.json
```diff
@@ -1,65 +1,80 @@
...
             "type": "object",
             "additionalProperties": false,
             "properties": {
-                "access_key": {"$ref": "#/definitions/key"},
-                "secret_key": {"$ref": "#/definitions/key"}
+                "access_key": {"$ref": "#/$defs/key"},
+                "secret_key": {"$ref": "#/$defs/key"},
+                "url": { "type": "string", "format": "uri" }
+            },
+            "if": {
```
sratcliffe: Is there a use key for URL with no keys ?

katsdpcontroller/schemas/s3_config.json
bmerry: Thanks, that's a typo. Fixed.

katsdpcontroller/schemas/s3_config.json
bmerry: Yes e.g. see the sandbox s3_config.json: we don't directly interact with the archive (yet) so we don't need any keys for it; but we need to know the read URL so that it can be embedded in telstate.

PR #457, opened 2020-01-15
Escape computed command-line arguments

bmerry:
Arguments in the `command` attribute of tasks are passed to `str.format`
to allow extra properties (mostly from the resolver) to be filled in at
resolution time. When a parameter is a variable rather than a literal,
there is a risk that it could contain a `{` which would break the
formatting. Variable string parameters are now passed to an escaping
function.

Closes SPR1-362.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -31,6 +31,11 @@ def normalise_gpu_name(name):
     return mangled
 
 
+def escape_format(s: str) -> str:
+    """Escape a string for :meth:`str.format`."""
+    return s.replace('{', '{{').replace('}', '}}')
+
+
 INGEST_GPU_NAME = 'GeForce GTX TITAN X'
 CAPTURE_TRANSITIONS = {
     CaptureBlockState.CAPTURING: [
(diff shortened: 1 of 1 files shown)


bmerry:
@sratcliffe this is low priority - we don't need it for spectral imager tests.

katsdpcontroller/generator.py
```diff
@@ -1778,17 +1783,19 @@ def _make_continuum_imager(g, config, capture_block_id, name, telstate, target_m
...
             '--secret-key', '{resolver.s3_config[continuum][read][secret_key]}',
-            '--select', f'scans="track"; corrprods="cross"; targets=[{target.description!r}]',
-            '--capture-block-id', capture_block_id,
-            '--output-id', name,
-            '--telstate-id', telstate.join(name, target_name),
-            '--outputdir', DATA_VOL.container_path,
-            '--mfimage', _render_continuum_parameters(mfimage_parameters),
-            '-w', '/mnt/mesos/sandbox', data_url
+            '--select',
+            escape_format(f'scans="track"; corrprods="cross"; targets=[{target.description!r}]'),
```
sratcliffe: Might be slightly more readable to construct a distinct list for the keys that want escaping, and then iterate through that to escape them.

bmerry: Yeah, I was thinking that. I'll definitely do it for the continuum imager. For the spectral imager the arguments of the two types are currently somewhat interleaved. I feel that sorting them by need for escaping will probably make the final command-line less readable (e.g. the `-i` arguments will no longer be grouped together), and there are fewer arguments that need escaping, so I'm tempted to leave things as is.

I also realised there were some bugs because I'm escaping URLs but they actually have genuine format strings in them (for the location of telstate) and instead `_stream_url` needs to do escaping. Will get back to it tomorrow.

bmerry:
@sratcliffe I think my reminder bot won't have poked you about this (because you've already left a review), but here's a reminder that it's been updated and ready for another review.

PR #458, opened 2020-01-15
Add counter for number of batch task retries

bmerry:
It's also added to the local copy of the imaging dashboard. The gauges
on the dashboard are also updated to give zero decimal places after the
counts (since they're integers).

### DIFF
--- a/katsdpcontroller/scheduler.py
+++ b/katsdpcontroller/scheduler.py
@@ -2380,6 +2380,9 @@ class TaskStats:
         """`n_tasks` batch tasks were skipped because a dependency failed."""
         pass
 
+    def batch_tasks_retried(self, n_tasks):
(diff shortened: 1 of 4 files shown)


katsdpcontroller/scheduler.py
```diff
@@ -2380,6 +2380,9 @@ def batch_tasks_skipped(self, n_tasks):
         """`n_tasks` batch tasks were skipped because a dependency failed."""
         pass
 
+    def batch_tasks_retried(self, n_tasks):
+        """`n_tasks` batch tasks failed and were re-scheduled."""
+
```
sratcliffe: Missing an explicit pass

bmerry: Thanks, added (it would actually have worked because the docstring functions like a `pass`, but it's inconsistent).

PR #463, opened 2020-03-13
Reduce capture-done timeout to fit within CAM timeout

sratcliffe:
The idea is to provide a sensible timeout length, but still fit comfortably inside the CAM timeout (400s) for this command. This will ensure that in a failure, CAM will be able to issue a deconfigure without hitting a busy state.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -42,7 +42,7 @@ CAPTURE_TRANSITIONS = {
         KatcpTransition('capture-init', '{capture_block_id}', timeout=30)
     ],
     CaptureBlockState.BURNDOWN: [
-        KatcpTransition('capture-done', timeout=360)
+        KatcpTransition('capture-done', timeout=240)
     ]
 }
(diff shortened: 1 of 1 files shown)


bmerry:
I'm a little rusty, but I think this timeout applies to each individual service that we issue a capture-done to, and there are dependency chains. So if one service takes 50 seconds, and the next times out after 360 seconds, then we're going to be at 410 seconds in total.

What was the timeout before you put in the temporary large value?

sratcliffe:
Aah - that would be an issue. It was 120s previously, but if this number is per-service, then we need a different strategy anyway, since the main aim is to avoid triggering a CAM led deconfigure whilst we are still busy...

bmerry:
Should we just put it back to 120s? It doesn't guarantee anything, but at least if there is only one service mis-behaving we have a good chance of timing out before CAM does.

sratcliffe:
@bmerry - 240s looks like it will be a happy middle ground - there are very few individual done's that take more than a few seconds...

bmerry:
Ok.

PR #465, opened 2020-03-25
Replace many instances of 'localhost' with '127.0.0.1'

bmerry:
There is variation in host systems as to whether localhost resolves to
::1 in addition to 127.0.0.1. This can cause problems if the system
doesn't also have Docker configured for IPv6.

This is mainly expected to affect developer test systems using the
sandbox. In particular, it doesn't address the problem of a
system-configured Mesos slave that advertises itself as 'localhost'.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -36,6 +36,9 @@ def escape_format(s: str) -> str:
     return s.replace('{', '{{').replace('}', '}}')
 
 
+# Docker doesn't support IPv6 out of the box, and on some systems 'localhost'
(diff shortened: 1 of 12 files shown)


bmerry:
@sratcliffe would you mind testing this on your machine after re-enabling the IPv6 localhost entry? Don't forget to build/pull the product controller image and push it to whatever image tag you're using so that you run the new product controller code.

Some of it is probably overkill (I think it only matters for connections where either the source or target is a container in non-host networking, which is relatively few), but even where it's not necessary it should ensure that behaviour is more consistent across systems.

PR #466, opened 2020-04-08
Abort when dependencies fail

bmerry:
When a realtime task fails, don't keep bumbling on running tasks that depend on it; kill them instead. This has a few benefits:
- Failures happen faster.
- There is less spam in the logs from tasks that start up, can't find the things they need in telstate, and vomit up errors that have nothing to do with the root cause.
- It sets up the groundwork for better error reporting back to CAM, since it'll be easier to figure out which tasks are the culprits (e.g. they won't have the death_expected flag set).

Relates to SPR1-424.

### DIFF
--- a/katsdpcontroller/product_controller.py
+++ b/katsdpcontroller/product_controller.py
@@ -10,7 +10,7 @@ import copy
 import uuid
 import functools
 from ipaddress import IPv4Address
-from typing import Dict, Set, List, Callable, Sequence, Optional, Type, Mapping
+from typing import Dict, Set, List, Tuple, Callable, Sequence, Optional, Type, Mapping
 
 import addict
 import jsonschema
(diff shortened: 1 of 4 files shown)


ludwigschwardt:
Ah, apoptosis! 😂 Been reading too much virial literature....

bmerry:
I've updated it to produce the better error message on failure: it now reports the tasks that died unexpectedly.

katsdpcontroller/product_controller.py
```diff
@@ -1141,10 +1151,13 @@ def cleanup(task):
                 await self._launch_telstate()
                 # launch containers for those nodes that require them
                 await self.sched.launch(self.physical_graph, self.resolver)
-                alive = self.check_nodes()
+                alive, died = self.check_nodes()
                 # is everything we asked for alive
                 if not alive:
-                    ret_msg = ("Some nodes in the graph failed to start. "
+                    fail_list = ', '.join(node.logical_node.name for node in died)
```
sratcliffe, line 1157: Why not `fail_list = ', '.join(node.logical_node.name for node in died) or 'Some nodes'` ?

katsdpcontroller/scheduler.py
```diff
@@ -2307,15 +2340,21 @@ def run_in_event_loop(func, *args, **kw):
...
+                    logger.info('Not starting %s because %s died', task.name, dep.name)
+                    raise aiohttp.web.HTTPServiceUnavailable(
+                        text=f'Dependency {dep.name} died without becoming ready\n')
+        except (asyncio.CancelledError, aiohttp.web.HTTPException):
+            raise
         except Exception as error:
             logger.exception('Exception while waiting for dependencies')
-            return aiohttp.web.HTTPInternalServerError(
-                reason='Exception while waiting for dependencies:\n{}\n'.format(error))
+            raise aiohttp.web.HTTPInternalServerError(
```
sratcliffe, line 2356: Does this get logged again by an outer handler ? We don't want the same error logged twice.

katsdpcontroller/product_controller.py
bmerry: Ok, done.

katsdpcontroller/scheduler.py
bmerry: The aiohttp logging will log the request URL and the 500 status, but it won't show the error message. And the container will print the error to stderr, but I don't think it'll generally get from there into our logging system.

PR #471, opened 2020-04-13
Spectral beam correction

bmerry:
Only enabled for L and UHF bands, since those are the only ones supported by katsdpimager.


katsdpcontroller/generator.py
```diff
@@ -1737,7 +1740,7 @@ def _get_targets(config, capture_block_id, name, telstate, min_time):
         else:
             logger.info('Skipping target %s: observed for %.1f seconds, threshold is %.1f',
                         target.name, obs_time, min_time)
-    return targets
+    return targets, data_set
```
mauch, line 1743: I'll be making some continuum imager changes that will require 'band awareness' in the not-too-distant future so this extra plumbing will come in handy there as well.

PR #472, opened 2020-04-14
Make unit tests reliable again

bmerry:
PR #466 changed the handling of failures of dependencies in a way that
required some mocking in the unit tests, but the way it was handled
wasn't reliable (basically, if the failure didn't propagate and set the
state to KILLING before the mocking code inspected the state, the
mocking code wouldn't set the state to DEAD). The approach to mocking
has been changed so that the KILLING -> DEAD transition is handled in a
wrapper around `dependency_abort`, which seems to be more reliable.


bmerry:
Looks like this isn't actually working yet (I fixed the test I'd previously broken, but broke others in the process). I'll close this PR and reopen once I've fixed it.

bmerry:
Was just a silly mistake - fixed now, and reopened the PR.

PR #473, opened 2020-04-20
Add depends_finished_critical edge property

bmerry:
When explicitly set to false, a depends_finished edge will force
ordering, but the downstream task will run even when the upstream task
fails. This is to support a future spectral imager report, which must
run after the result of spectral imaging but should run even if
individual spectral tasks fail.

Relates to SPR1-108.

### DIFF
--- a/katsdpcontroller/scheduler.py
+++ b/katsdpcontroller/scheduler.py
@@ -204,6 +204,7 @@ DEPENDS_RESOURCES = 'depends_resources'
 DEPENDS_RESOLVE = 'depends_resolve'
 DEPENDS_KILL = 'depends_kill'
 DEPENDS_FINISHED = 'depends_finished'      # for batch tasks
+DEPENDS_FINISHED_CRITICAL = 'depends_finished_critical'
 DECIMAL_CONTEXT = decimal.Context(traps=[
     decimal.Overflow, decimal.InvalidOperation, decimal.DivisionByZero,  # defaults
     decimal.Inexact, decimal.FloatOperation])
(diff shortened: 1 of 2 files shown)


katsdpcontroller/scheduler.py
```diff
@@ -3296,9 +3300,13 @@ async def run(node_set):
...
                                 desc = node.name + ' (and {} others)'.format(len(node_set) - 1)
-                            logger.info('Skipping %s because %s failed', desc, dep.name)
-                            self.task_stats.batch_tasks_skipped(len(node_set))
-                            raise TaskSkipped(node) from None
+                            if critical:
+                                logger.info('Skipping %s because %s failed', desc, dep.name)
+                                self.task_stats.batch_tasks_skipped(len(node_set))
+                                raise TaskSkipped(node) from None
+                            else:
+                                logger.debug('Continuing with %s after non-critical %s failed',
```
sratcliffe, line 3308: Torn a bit here - I feel that this should actually be promoted all the way to WARNING. The logic is that I would always like a WARNING level event when something occurs that could lead to output data or products not being what you would expect. In this case for instance the report could be missing channels. 

I realise that the failure event leading here in the first place will log useful things... And this is probably overthinking a simple log message...

bmerry: > I realise that the failure event leading here in the first place will log useful things

That was my thinking - having a higher log level here is just going to log everything twice, and the first failure will have more useful information. I also aim to have the report itself indicate channels for which no statistics arrived (for channels where everything is flagged I'll still have the pipeline store something to indicate that, so that it can be distinguished from failures).

sratcliffe: Fair enough

PR #475, opened 2020-05-05
Run spectral imager report

bmerry:
The spectral imager pipeline now also takes an extra parameter so that it can extract the appropriate information from telstate. This PR should thus be merged at the same time as ska-sa/katsdpimager#38.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -1826,6 +1826,7 @@ def _make_spectral_imager(g, config, capture_block_id, name, telstate, target_ma
     band = data_set.spectral_windows[data_set.spw].band if data_set is not None else ''
     del data_set    # Allow Python to recover the memory
 
+    nodes = []
     for target, obs_time in targets:
         for i in range(0, l0_info.n_channels, SPECTRAL_OBJECT_CHANNELS):
             first_channel = max(i, output_channels[0])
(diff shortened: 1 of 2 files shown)


katsdpcontroller/generator.py
```diff
@@ -1899,6 +1902,31 @@ def _make_spectral_imager(g, config, capture_block_id, name, telstate, target_ma
...
+    # list to allow for applying a static mask in future.
+    view['output_channels'] = list(range(output_channels[0], output_channels[1]))
+    return data_url, nodes
+
+
+def _make_spectral_imager_report(g, config, capture_block_id, name, data_url, spectral_nodes):
+    report = SDPLogicalTask(f'spectral_image_report.{name}')
+    report.cpus = 1.0
+    # Memory is a guess - but since we don't open the chunk store it should be lightweight
+    report.mem = 4 * 1024
```
sratcliffe, line 1915: Happy enough with a guess, but given the severity of OOM, is there any scaling in the report related to the field complexity or observation length ? (I doubt it, but just thought I would check)

bmerry: I do use katdal is used to open the observation. However, I don't access the data at all, and it's a live telstate so it won't need to copy all the raw sensor data into memory. There will certainly be some scaling with time and with channels, but it should be small since it won't scale with time × channels × baselines, just time + channels.

bmerry:
Thanks. I've got a slot on Thursday where I'll run off this branch to see how it looks.

PR #477, opened 2020-05-27
Extend config schema with spectral imager parameters

bmerry:
This allows overriding settings like the major cycle gain. Not all
parameters to katsdpimager are allowed e.g. those that would affect the
pixel size or FOV are disallowed because that would impact storage
requirements.


bmerry:
I need this to tweak the imager parameters when doing short simulations with bad PSFs. Once approved I'll update the ICD documentation.

bmerry:
> Now that we are getting a significant number of these parameters, I wonder if we should look at ways of either auto-generating these from the underlying package that supports the parameters, or at least having some version lock to prevent the command-line interface drifting from this specification.

Given that we've had a similar thing for cal parameters for years and it's never caused issues (most likely because no one ever uses it), I'm not too worried. I'll worry about it more if CAM ever provide a way for users to actually supply config parameters, and the users start using it.

bmerry:
I realised that the cal parameters used the underscore-ised form of the parameter name, so I changed the spectral imager parameters to match.

PR #478, opened 2020-05-29
Bumble on with postprocessing

bmerry:
If a capture-done fails, or one of the postprocessing transitions fails
(which I think is just the capture-done for the flag writer), carry on
anyway with postprocessing.

Fixes SPR1-358. It possibly introduces much worse issues in the corner
cases, particularly if cal jams up while writing the report and hence
never finishes with the capture block.


bmerry:
Ok, it's merged. I'll leave it up to you to schedule deployment, since it's probably you that will get the support call if it all goes wrong, or the lack of support calls if it does what it's supposed to do.

PR #480, opened 2020-06-05
Send X-Poweroff-Server header to poweroff_server

bmerry:
This is needed after ska-sa/poweroff_server#2.

### DIFF
--- a/katsdpcontroller/test/test_master_controller.py
+++ b/katsdpcontroller/test/test_master_controller.py
@@ -884,8 +884,8 @@ class TestDeviceServer(asynctest.ClockedTestCase):
         await self.client.request('capture-init', 'product')
         reply, informs = await self.client.request('sdp-shutdown')
         self.assertEqual(reply[0], b'127.0.0.144,127.0.0.42')
-        poweroff_mock.assert_any_call(url1, data=None)
-        poweroff_mock.assert_any_call(url2, data=None)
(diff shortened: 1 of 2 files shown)


bmerry:
Marking as draft for now because I still need to test it on Monday.

bmerry:
Have tested, so now ready for review.

katsdpcontroller/test/test_master_controller.py
```diff
@@ -884,8 +884,8 @@ async def test_sdp_shutdown(self, rmock: aioresponses) -> None:
         await self.client.request('capture-init', 'product')
         reply, informs = await self.client.request('sdp-shutdown')
         self.assertEqual(reply[0], b'127.0.0.144,127.0.0.42')
-        poweroff_mock.assert_any_call(url1, data=None)
-        poweroff_mock.assert_any_call(url2, data=None)
+        poweroff_mock.assert_any_call(url1, headers={'X-Poweroff-Server': '1'}, data=None)
```
sratcliffe, line 887: Is it worth testing without the X-Poweroff-Server header set ?

bmerry: This is a mock, so I don't see what the point would be - it would just confirm that when we don't pass the header, the header isn't there.

It would be a sensible test on the poweroff_server repo, but that doesn't currently have any unit tests.

sratcliffe: Fair enough...

PR #482, opened 2020-06-29
Remove support for 1.x product-configure schema

bmerry:
2.x has been around for a long time now and is described in the ICD, so
there is little reason to hang on to 1.x, and getting rid of it will
reduce the complexity of a 3.x I want to introduce.

### DIFF
--- a/katsdpcontroller/schemas/product_config.json.j2
+++ b/katsdpcontroller/schemas/product_config.json.j2
@@ -6,19 +6,13 @@
     "properties": {
         "version": {
             "type": "string",
-            "enum": ["1.0", "1.1", "2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6"]
+            "enum": ["2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6"]
         }
     }
 }
 {% endmacro %}
 
 {% macro validate(version) %}
-{% if version >= "2.0" %}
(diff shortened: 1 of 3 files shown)


ajoubertza:
Fine with me.  CAM is using schema version 2.4 (or at least claims to!)

https://github.com/ska-sa/katproxy/blob/0b3d20db7bddc5b23ab8a8f73c7e1e31e9743f2b/katproxy/proxy/data_sdp_manager.py#L673

katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -6,19 +6,13 @@
     "properties": {
         "version": {
             "type": "string",
-            "enum": ["1.0", "1.1", "2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6"]
```
sratcliffe, line 9: Any impact to retiring 1.x support ? (I suspect not, but need to ask)

sratcliffe: Happy with that - if we are going to cull we probably should cull to the latest version...

bmerry: I'll remove pre-2.0 for now. Trying to remove 2.0-2.3 in this PR will probably conflict with another branch I'm working on, but I'll take this as license to drop support for them in a later PR.

PR #483, opened 2020-07-03
Implement product-config schema version 3

bmerry:
The new schema is documented at https://docs.google.com/document/d/15ZIyuBf4Vk4ESIEdKsQxdSBoLEwh85eyKeDJouLRAZY/edit#heading=h.v512w2fvji5f.

The implementation is greatly changed, using Python classes to wrap the various streams. These classes combine the functionality of semantic validation and the \*Info classes that were in generator.py, and avoid a lot of tedious dict lookups. There is also lots of type annotations so that it's harder to get confused about which class you're using and look up the wrong thing.

There are also a number of bugs that I noticed and fixed along the way:
- The code for getting sensor values from katportal to override the values in the config was not getting called (SPR1-594).
- The rate_ratio config value was never used (the default of 8 was always used regardless).
- The calculation of cal output bandwidth was wrong (it assumed exactly one flags stream with no spectral averaging)

To keep things slightly simpler, schema versions 2.0-2.3 have been dropped, which basically just means some validation code could be removed since newer versions are strictly more general.

One regression is that there is now no way to specify a single-pol CBF. However, the internal support is still there, and would just require a way to detect such an instrument from sensors (e.g. from the baseline list), and would not require a schema change.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -14,7 +14,7 @@ import numpy as np
 import katdal
 import katdal.datasources
 import katpoint
-from katsdptelstate.endpoint import Endpoint, endpoint_list_parser
+from katsdptelstate.endpoint import Endpoint
 
 from . import scheduler, product_config
 from .tasks import (
@@ -271,7 +271,6 @@ def _make_cbf_simulator(g: networkx.MultiDiGraph,
     ibv = not configuration.options.develop
 
     def make_cbf_simulator_config(task, resolver):
-        substreams = stream.n_channels // stream.n_channels_per_substream
         conf = {
             'cbf_channels': stream.n_channels,
             'cbf_adc_sample_rate': stream.adc_sample_rate,
@@ -295,8 +294,7 @@ def _make_cbf_simulator(g: networkx.MultiDiGraph,
                 'beamformer_timesteps': stream.spectra_per_heap,
                 'beamformer_bits': stream.out_bits_per_sample
             })
-        if 'center_freq' in settings:
-            conf['cbf_center_freq'] = float(stream.centre_frequency)
+        conf['cbf_center_freq'] = float(stream.centre_frequency)
         conf['cbf_antennas'] = [{'description': str(ant)} for ant in stream.antenna_objects]
         return conf
 
@@ -444,9 +442,8 @@ def _make_timeplot_beamformer(
         stream: product_config.TiedArrayChannelisedVoltageStreamBase) -> \
             Tuple[scheduler.LogicalNode, scheduler.LogicalNode]:
     """Make timeplot server for the beamformer, plus a beamformer capture to feed it."""
-    develop = configuration.options.develop
     beamformer = _make_beamformer_engineering_pol(
-        g, None, stream, f'bf_ingest_timeplot.{stream.name}', False, 0, develop)
+        g, configuration, None, stream, f'bf_ingest_timeplot.{stream.name}', False, 0)
     beamformer.critical = False
 
     # It's a low-demand setup (only one signal). The CPU and memory numbers
@@ -529,43 +526,6 @@ def n_cal_nodes(configuration: Configuration,
         return 1
     else:
         return 4
-
-
-def _adjust_ingest_output_channels(config, names):
-    """Modify the config dictionary to set output channels for ingest.
-
-    It is widened where necessary to meet alignment requirements.
-
-    Parameters
-    ----------
-    config : dict
-        Configuration dictionary
-    names : list of str
-        The names of the ingest output products. These must match i.e. be the
-        same other than for continuum_factor.
-    """
-    def lcm(a, b):
-        return a // math.gcd(a, b) * b
-
-    n_ingest = n_ingest_nodes(config, names[0])
-    outputs = [config['outputs'][name] for name in names]
-    src = config['inputs'][outputs[0]['src_streams'][0]]
-    acv = config['inputs'][src['src_streams'][0]]
-    n_chans = acv['n_chans']
-    requested = outputs[0]['output_channels']
-    alignment = 1
-    # This is somewhat stricter than katsdpingest actually requires, which is
-    # simply that each ingest node is aligned to the continuum factor.
-    for output in outputs:
-        alignment = lcm(alignment, n_ingest * output['continuum_factor'])
-    assigned = [_round_down(requested[0], alignment), _round_up(requested[1], alignment)]
-    # Should always succeed if product_config.validate passed
-    assert 0 <= assigned[0] < assigned[1] <= n_chans, "Aligning channels caused an overflow"
-    for name, output in zip(names, outputs):
-        if assigned != requested:
-            logger.info('Rounding output channels for %s from %s to %s',
-                        name, requested, assigned)
-        output['output_channels'] = assigned
 
 
 def _make_ingest(g: networkx.MultiDiGraph, configuration: Configuration,
@@ -1074,23 +1034,20 @@ def _make_imager_writers(g: networkx.MultiDiGraph,
                          stream: product_config.ImageStream,
                          max_channels: Optional[int] = None) -> None:
     """Make vis and flag writers for an imager output"""
-    orig_flags_name = config['outputs'][name]['src_streams'][0]
-    orig_l0_name = config['outputs'][orig_flags_name]['src_streams'][0]
-    cal = config['outputs'][orig_flags_name]['calibration'][0]
     _make_vis_writer(g, configuration, stream.vis, s3_name=s3_name,
                      local=False, prefix=stream.name, max_channels=max_channels)
     _make_flag_writer(g, configuration, stream.flags, s3_name=s3_name,
-                      local=False, prefix=name, max_channels=max_channels)
+                      local=False, prefix=stream.name, max_channels=max_channels)
 
 
 def _make_beamformer_engineering_pol(
-    g: networkx.MultiDiGraph,
-    stream: Optional[product_config.BeamformerEngineeringStream],
-    src_stream: product_config.TiedArrayChannelisedVoltageStreamBase,
-    node_name: str,
-    ram: bool,
-    idx: int,
-    develop: bool) -> SDPLogicalTask:
+        g: networkx.MultiDiGraph,
+        configuration: Configuration,
+        stream: Optional[product_config.BeamformerEngineeringStream],
+        src_stream: product_config.TiedArrayChannelisedVoltageStreamBase,
+        node_name: str,
+        ram: bool,
+        idx: int) -> SDPLogicalTask:
     """Generate node for a single polarisation of beamformer engineering output.
 
     This handles two cases: either it is a capture to file, associated with a
@@ -1100,6 +1057,10 @@ def _make_beamformer_engineering_pol(
 
     Parameters
     ----------
+    g
+        Graph under construction.
+    configuration
+        Product configuration.
     stream
         The beamformer engineering stream, if applicable. If ``None``, this is
         for signal displays.
@@ -1108,9 +1069,9 @@ def _make_beamformer_engineering_pol(
     node_name
         Name to use for the logical node
     ram
-        Whether this node is for writing to ramdisk (ignored if `timeplot` is true)
+        Whether this node is for writing to ramdisk (ignored if `stream` is ``None``)
     idx
-        Number of this source within the output (ignored if `timeplot` is true)
+        Number of this source within the output (ignored if `stream` is ``None``)
     develop
         Whether this is a develop-mode config
     """
@@ -1133,7 +1094,7 @@ def _make_beamformer_engineering_pol(
         # process, so we need more memory allocation than there is
         # space in the ramdisk. This is only used for lab testing, so
         # we just hardcode a number.
-        bf_ingest.ram = 220 * 1024
+        bf_ingest.mem = 220 * 1024
     # In general we want it on the same interface as the NIC, because
     # otherwise there is a tendency to drop packets. But for ramdisk capture
     # this won't cut it because the machine for this is dual-socket and we
@@ -1141,7 +1102,7 @@ def _make_beamformer_engineering_pol(
     # regions.
     bf_ingest.interfaces = [
         scheduler.InterfaceRequest('cbf',
-                                   infiniband=not develop,
+                                   infiniband=not configuration.options.develop,
                                    affinity=timeplot or not ram)
     ]
     # XXX Even when there is enough bandwidth, sharing a node with correlator
@@ -1163,7 +1124,7 @@ def _make_beamformer_engineering_pol(
         config = {
             'affinity': [task.cores['disk'], task.cores['network']],
             'interface': task.interfaces['cbf'].name,
-            'ibv': not develop,
+            'ibv': not configuration.options.develop,
             'stream_name': src_stream.name,
             'aiomonitor': True
         }
@@ -1193,7 +1154,10 @@ def _make_beamformer_engineering_pol(
     return bf_ingest
 
 
-def _make_beamformer_engineering(g, config, name):
+def _make_beamformer_engineering(
+        g: networkx.MultiDiGraph,
+        configuration: Configuration,
+        stream: product_config.BeamformerEngineeringStream) -> Sequence[scheduler.LogicalTask]:
     """Generate nodes for beamformer engineering output.
 
     If `timeplot` is true, it generates services that only send data to
@@ -1201,17 +1165,12 @@ def _make_beamformer_engineering(g, config, name):
     refer to an ``sdp.beamformer`` rather than an
     ``sdp.beamformer_engineering`` stream.
     """
-    output = config['outputs'][name]
-    srcs = output['src_streams']
-    ram = output.get('store') == 'ram'
-    develop = is_develop(config)
-
+    ram = stream.store == 'ram'
     nodes = []
-    for i, src in enumerate(srcs):
-        info = BeamformerInfo(config, name, src)
+    for i, src in enumerate(stream.tied_array_channelised_voltage):
         nodes.append(_make_beamformer_engineering_pol(
-            g, stream, stream.tied_array_channelised_voltage,
-            'bf_ingest.{}.{}'.format(name, i + 1), src, ram, i, develop))
+            g, configuration, stream, src,
+            'bf_ingest.{}.{}'.format(stream.name, i + 1), ram, i))
     return nodes
 
 
@@ -1251,8 +1210,8 @@ def build_logical_graph(configuration: Configuration,
                 input_multicast.append(node)
 
     # cam2telstate node (optional: if we're simulating, we don't need it)
-    cam_http = configuration.by_class(product_config.CamHttpStream)
     cam2telstate: Optional[scheduler.LogicalNode] = None
+    cam_http = configuration.by_class(product_config.CamHttpStream)
     if cam_http:
         cam2telstate = _make_cam2telstate(g, configuration, cam_http[0])
         for node in input_multicast:
@@ -1273,13 +1232,14 @@ def build_logical_graph(configuration: Configuration,
         if stream.continuum_factor == 1:
             for stream2 in configuration.by_class(product_config.VisStream):
                 if (stream2 not in l0_done and stream2.continuum_factor > 1
-                    and stream2.compatible(stream)):
-                        _adjust_ingest_output_channels(config, [name, name2])  # TODO
-                        _make_ingest(g, configuration, stream, stream2)
-                        _make_timeplot_correlator(g, configuration, stream)
-                        l0_done.add(stream)
-                        l0_done.add(stream2)
-                        break
+                        and stream2.compatible(stream)):
+                    # TODO: restore this functionality
+                    # _adjust_ingest_output_channels(config, [name, name2])
+                    _make_ingest(g, configuration, stream, stream2)
+                    _make_timeplot_correlator(g, configuration, stream)
+                    l0_done.add(stream)
+                    l0_done.add(stream2)
+                    break
 
     l0_spectral_only = False
     l0_continuum_only = False
@@ -1289,7 +1249,8 @@ def build_logical_graph(configuration: Configuration,
             l0_spectral_only = True
         else:
             l0_continuum_only = True
-        _adjust_ingest_output_channels(configuration, [stream])
+        # TODO: restore this functionality
+        # _adjust_ingest_output_channels(configuration, [stream])
         if is_spectral:
             _make_ingest(g, configuration, stream, None)
             _make_timeplot_correlator(g, configuration, stream)
@@ -1304,7 +1265,6 @@ def build_logical_graph(configuration: Configuration,
             _make_vis_writer(g, configuration, stream, 'archive', local=True)
 
     for stream in configuration.by_class(product_config.CalStream):
-        src = stream.vis
         # Check for all corresponding flags outputs
         flags = []
         for flags_stream in configuration.by_class(product_config.FlagsStream):
@@ -1387,15 +1347,15 @@ def build_logical_graph(configuration: Configuration,
     return g
 
 
-def _continuum_imager_cpus(config):
-    return 24 if not is_develop(config) else 2
+def _continuum_imager_cpus(configuration: Configuration):
+    return 24 if not configuration.options.develop else 2
 
 
-def _spectral_imager_cpus(config):
+def _spectral_imager_cpus(configuration: Configuration):
     # Fairly arbitrary number, based on looking at typical usage during a run.
     # In practice the number of spectral imagers per box is limited by GPUs,
     # so the value doesn't make a huge difference.
-    return 2 if not is_develop(config) else 1
+    return 2 if not configuration.options.develop else 1
 
 
 def _stream_url(capture_block_id, stream_name):
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -90,6 +90,8 @@ CONFIG_VOL = scheduler.VolumeRequest('config', '/var/kat/config', 'RW')
 WRITER_OBJECT_SIZE = 20e6    # 20 MB
 #: Maximum channels per chunk for spectral imager
 SPECTRAL_OBJECT_CHANNELS = 128
+
+# TODO: remove these
 #: Minimum observation time for continuum imager (seconds)
 DEFAULT_CONTINUUM_MIN_TIME = 15 * 60     # 15 minutes
 #: Minimum observation time for spectral imager (seconds)
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -22,7 +22,7 @@ from .tasks import (
     SDPLogicalTask, SDPPhysicalTask, LogicalGroup,
     CaptureBlockState, KatcpTransition)
 from .product_config import (
-    BYTES_PER_VIS, BYTES_PER_FLAG, BYTES_PER_VFW, bandwidth,
+    BYTES_PER_VIS, BYTES_PER_FLAG, BYTES_PER_VFW, data_rate,
     Configuration)
 # Import just for type annotations, but avoid at runtime due to circular imports
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        'sim.cbf.baseline_correlation_products'
+    }
+
+    def __init__(self, name: str, src_streams: Sequence[Stream], *,
+                 int_time: float,
+                 output_channels: Optional[Tuple[int, int]] = None,
+                 continuum_factor: int,
+                 excise: bool,
+                 archive: bool,
+                 n_servers: int) -> None:
```
sratcliffe, line 780: Could you call this n_ingest_servers just to be clear ?

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+            continuum_factor=config['continuum_factor'],
+            excise=config.get('excise', True),
+            archive=config['archive'],
+            n_servers=4 if not options.develop else 2
+        )
+
+    def compatible(self, other: 'VisStream') -> bool:
+        """Determine whether the configurations are mostly the same.
+
+        Specifically, they must be the same other than the vlaues of
```
sratcliffe, line 867: vlaues -> values

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        ``continuum_factor`` and ``archive``.
+        """
+        return (
```
sratcliffe, line 870: wouldn't it be more robust to make copies of self and other, remove continuum_factor and archive and then compare them directly ? Saves updating this section if you add new things.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+    stream_type = 'sdp.beamformer_engineering'
+
+    def __init__(self, name: str, src_streams: Sequence[Stream], *,
+                 store: str,
+                 output_channels: Optional[Tuple[int, int]] = None) -> None:
+        super().__init__(name, src_streams)
+        cbf_channels = self.antenna_channelised_voltage.n_channels
+        c = _normalise_output_channels(cbf_channels, output_channels)
+        for tacv in self.tied_array_channelised_voltage:
+            c = _normalise_output_channels(cbf_channels, c, tacv.n_channels_per_endpoint)
```
sratcliffe, line 936: Not sure why these need to be normalised per endpoint, but will ask you over video about this.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+                stream_cls = STREAM_CLASSES[stream_config['type']]
+                for sensor in stream_cls._class_sensors:
+                    instrument = stream_config['instrument_dev_name']
+                    full_name = sensor.full_name(components, name, instrument)
+                    try:
+                        sample = await client.sensor_value(full_name)
+                    except Exception as exc:
+                        raise SensorFailure(f'Could not get value for {full_name}: {exc}') from exc
+                    if sample.status not in {'nominal', 'warn', 'error'}:
+                        raise SensorFailure(f'Sensor {full_name} has status {sample.status}')
```
sratcliffe, line 1244: very minor: "...has invalid status..."

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        # the cam.http stream, but that is harmless.
+        for name in networkx.topological_sort(g):
+            stream_config = stream_configs[name]
+            stream_cls = STREAM_CLASSES[stream_config['type']]
+            src_streams = [streams[d] for d in stream_config.get('src_streams', [])]
+            try:
+                streams[name] = stream_cls.from_config(
+                    options, name, stream_config, src_streams, sensors[name])
+            except ValueError as exc:
+                raise ValueError(f'Configuration error for {name}: {exc}') from exc
```
sratcliffe, line 1271: minor: "...error for stream {name}..."

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+            if stream_config['type'] == 'cam.http':
+                cam_http = CamHttpStream.from_config(options, name, stream_config, [], {})
+                break
+
+        # Extract sensor values. This is pulled into a separate block rather
+        # than done as we construct each stream so that one day it can be
+        # optimised to make a single sensor_values call for all sensors.
+        sensors: Dict[str, Dict[str, Any]] = {name: {} for name in stream_configs}
+        if cam_http:
+            client = katportalclient.KATPortalClient(str(cam_http.url), None)
```
sratcliffe, line 1225: Does the client not need any disposal once we are done with it ?

katsdpcontroller/product_config.py
```diff
@@ -1,22 +1,73 @@
...
 import katportalclient
 
 from . import schemas
 
 
 logger = logging.getLogger(__name__)
+_S = TypeVar('_S', bound='Stream')
+_ValidTypes = Union[AbstractSet[str], Sequence[str]]
+#: Minimum observation time for continuum imager (seconds)
+DEFAULT_CONTINUUM_MIN_TIME = 15 * 60.0     # 15 minutes
```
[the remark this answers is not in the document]
bmerry: I don't mind making a `defaults.py` to hold things like this. It's not totally clear to me which things should move there though. My candidates would be:

```python
INGEST_GPU_NAME = 'GeForce GTX TITAN X'
#: Maximum number of custom signals requested by (correlator) timeplot
TIMEPLOT_MAX_CUSTOM_SIGNALS = 256
#: Target size of objects in the object store
WRITER_OBJECT_SIZE = 20e6    # 20 MB
#: Maximum channels per chunk for spectral imager
SPECTRAL_OBJECT_CHANNELS = 128
#: Minimum observation time for continuum imager (seconds)
DEFAULT_CONTINUUM_MIN_TIME = 15 * 60.0     # 15 minutes
#: Minimum observation time for spectral imager (seconds)
DEFAULT_SPECTRAL_MIN_TIME = 3600.0         # 1 hour
#: Size of cal buffer in seconds
DEFAULT_CAL_BUFFER_TIME = 25 * 60.0        # 25 minutes (allows a single batch of 15 minutes)
#: Maximum number of scans to include in report
DEFAULT_CAL_MAX_SCANS = 1000
#: Speed at which flags are transmitted, relative to real time
DEFAULT_FLAGS_RATE_RATIO = 8.0
#: Alignment constraint for `int_time` in katcbfsim
KATCBFSIM_SPECTRA_PER_HEAP = 256
```

Can you look though generator.py and product_config.py and let me know if there are others you'd want to move?

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        raise ValueError(f'output_channels is empty ({c[0]}:{c[1]})')
+    elif c[0] < 0 or c[1] > n_channels:
+        raise ValueError(f'output_channels ({c[0]}:{c[1]}) overflows valid range 0:{n_channels}')
+    else:
+        return (c[0] // alignment * alignment,
+                (c[1] + alignment - 1) // alignment * alignment)
+        return c
+
+
+def bandwidth(size: float, time: float, ratio: float = 1.05, overhead: float = 128) -> float:
```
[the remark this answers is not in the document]
bmerry: I like that - wish I'd thought of that when I first wrote the Info classes and called things "net_bandwidth". Fixed.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        self.sources = list(sources)
+
+    @classmethod
+    def from_config(cls, config: Mapping[str, Any]) -> 'Simulation':
+        sources = []
+        for i, desc in enumerate(config.get('sources', []), 1):
+            try:
+                source = katpoint.Target(desc)
+            except Exception as exc:
+                raise ValueError(f'Invalid source {i}: {exc}') from exc
```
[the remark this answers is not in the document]
bmerry: So that you can go back to the catalogue in the configuration and figure out which source is the problem.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        Time between heaps in seconds
+    ratio
+        Relative overhead
+    overhead
+        Absolute overhead, in bytes
+    """
+    return (size * ratio + overhead) * 8 / time
+
+
+class ServiceOverride:
```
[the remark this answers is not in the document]
bmerry: Fair enough, since I can't find a description in the ICD. I've added a docstring.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        self.adc_sample_rate = adc_sample_rate
+        self.n_samples_between_spectra = n_samples_between_spectra
+
+
+class AntennaChannelisedVoltageStream(CbfStream, AntennaChannelisedVoltageStreamBase):
+    """Real antenna-channelised-voltage stream."""
+
+    stream_type: ClassVar[str] = 'cbf.antenna_channelised_voltage'
+    _class_sensors: ClassVar[Sequence[_Sensor]] = [
+        _CBFSensor('n_chans', int),
```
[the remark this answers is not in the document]
bmerry: It's not possible any more to be 100% consistent everywhere because the SDP-CAM ICD uses "output_channels" rather than "output_chans", but I've renamed all instances of n_channels and n_channels_per_x to n_chans and n_chans_per_x to be a little less inconsistent.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+
+    @classmethod
+    def from_config(cls,
+                    options: Options,
+                    name: str,
+                    config: Mapping[str, Any],
+                    src_streams: Sequence['Stream'],
+                    sensors: Mapping[str, Any]) -> 'SimAntennaChannelisedVoltageStream':
+        antennas = []
+        for desc in config['antennas']:
```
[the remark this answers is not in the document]
bmerry: In that case we have only the antenna name, not a description.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+                f'n_channels ({self.n_channels}) is not '
+                f'a multiple of endpoints ({self.n_endpoints})')
+        if self.n_channels_per_endpoint % self.n_channels_per_substream != 0:
+            raise ValueError(
+                f'channels per endpoint ({self.n_channels_per_endpoint}) '
+                f'is not a multiple of channels per substream ({self.n_channels_per_substream})'
+            )
+
+    @property
+    def antenna_channelised_voltage(self) -> 'AntennaChannelisedVoltageStreamBase':
```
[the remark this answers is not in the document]
bmerry: If I do that then any function that modifies `src_streams` will also need to remember to modify `antenna_channelised_voltage` (and if it's done in code that is working on the base class, it will need separate cases for each subclass). While I don't believe there is any such code at the moment, I prefer not to duplicate information for this reason.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+        super().__init__(
+            name, src_streams,
+            n_endpoints=n_endpoints,
+            n_channels_per_substream=n_channels_per_substream,
+            bits_per_sample=bits_per_sample
+        )
+        self._int_time = int_time
+        self.n_baselines = n_baselines
+
+    @property
```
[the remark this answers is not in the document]
bmerry: In this case it's because `int_time` is an abstract property of the base class. I don't want to make it an attribute of the base class because in TiedArrayChannelisedVoltageStream it's a function of the other parameters and cannot be set independently.

katsdpcontroller/product_config.py
bmerry: I'm reluctant because at some point I will probably want to give CalStream the same property and maybe even share some code to validate that the number of servers divides into the number of channels, and that's easier if they have a common attribute.

katsdpcontroller/product_config.py
bmerry: Thanks, fixed.

katsdpcontroller/product_config.py
bmerry: Yes, I wasn't really happy with this, but thought that trying to compare two streams for equality would still require listing all the attributes. But this is Python, so I've solved it with some reflection.

katsdpcontroller/product_config.py
bmerry: For the very hypothetical case that the two streams have different numbers of endpoints, you'd want to align the output channel range to the coarser of the two (actually it would need to be aligned to the LCM of the two, but as long as we're dealing with power-of-two channelisation the larger value will always be a multiple of the smaller one).

katsdpcontroller/product_config.py
bmerry: I'm not sure I'd call other statuses "invalid": it's there are other statuses that are valid in katcp, they just indicate that the value isn't usable.

katsdpcontroller/product_config.py
bmerry: Done.

katsdpcontroller/product_config.py
bmerry: I looked into that. There is a `disconnect` method, but that's only meaningful if it's established a websocket connection for streaming updates. For HTTP requests it uses a tornado AsyncHTTPClient, which has a singleton per IOLoop, so doesn't need cleanup.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+            buffer_time=config.get('buffer_time', DEFAULT_CAL_BUFFER_TIME),
+            max_scans=config.get('max_scans', DEFAULT_CAL_MAX_SCANS)
+        )
+
+
+class FlagsStream(Stream):
+    """An instance of sdp.flags."""
+
+    stream_type: ClassVar[str] = 'sdp.flags'
+    _valid_src_types: ClassVar[_ValidTypes] = ['sdp.vis', 'sdp.cal']
```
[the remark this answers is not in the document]
bmerry: A list is matched up positionally i.e. first element must be sdp.vis, second element must be sdp.cal. I've added a comment to `Stream` to document this.

katsdpcontroller/product_config.py
```diff
@@ -183,152 +1318,60 @@ def validate(config):
...
-                            n_chans, n_endpoints))
-                n_chans_per_endpoint = n_chans // n_endpoints
-                if n_chans_per_endpoint % n_chans_per_substream != 0:
-                    raise ValueError(
-                        'channels per endpoints ({}) not a multiple of n_chans_per_substream ({})'
-                        .format(n_chans_per_endpoint, n_chans_per_substream))
-        except ValueError as error:
-            raise ValueError('{}: {}'.format(name, error)) from error
+                raise ValueError(f'Unknown source {src} in {name}')
+            if isinstance(valid_types, collections.abc.Set):
```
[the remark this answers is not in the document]
bmerry: Most classes have a set, although most of them only have a single-element set. VisStream has a non-trivial set.

katsdpcontroller/product_config.py
```diff
@@ -183,152 +1318,60 @@ def validate(config):
...
+                valid_type = src_config['type'] in valid_types
+            else:
+                valid_type = src_config['type'] == valid_types[i]
+            if not valid_type:
+                raise ValueError(f'Source {src} has wrong type for {name}')
+
+    have_cam_http = False
+    for name, stream in inputs.items():
+        # It's not possible to convert 2.x simulations to 3.0 because we don't
```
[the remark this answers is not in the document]
bmerry: It would offend me :-) It would also make this function more complicated, for little benefit since it's easy enough to upgrade kattelmod to produce v3 config files.

katsdpcontroller/generator.py
```diff
@@ -636,55 +408,62 @@ def _make_timeplot(g, name, description,
...
+def _make_timeplot_correlator(g: networkx.MultiDiGraph,
+                              configuration: Configuration,
+                              stream: product_config.VisStream) -> scheduler.LogicalNode:
+    n_ingest = stream.n_servers
 
     # Exact requirement not known (also depends on number of users). Give it
     # 2 CPUs (max it can use) for 16 antennas, 32K channels and scale from there.
     # Lower bound (from inspection) to cater for fixed overheads.
-    cpus = max(2 * min(1.0, spectral_info.n_vis / _N16_32), 0.3)
+    cpus = max(2 * min(1.0, stream.n_baselines * stream.n_spectral_channels / _N16_32), 0.3)
```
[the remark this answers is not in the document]
bmerry: Are you suggesting adding `stream.n_spectral_vis`?

katsdpcontroller/generator.py
```diff
@@ -700,31 +479,32 @@ def _correlator_timeplot_frame_size(spectral_info, n_cont_channels, n_ingest):
...
 
-def _beamformer_timeplot_bandwidth(info):
+def _beamformer_timeplot_bandwidth(
+        stream: product_config.TiedArrayChannelisedVoltageStreamBase) -> float:
     """Bandwidth for the beamformer timeplot stream.
 
     Parameters
     ----------
-    info : :class:`TiedArrayChannelisedVoltageInfo`
+    info
```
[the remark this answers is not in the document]
bmerry: Fixed.

katsdpcontroller/product_config.py
```diff
@@ -31,63 +82,1196 @@ class _Sensor(ABC):
...
+    def from_config(cls: Type[_S],
+                    options: Options,
+                    name: str,
+                    config: Mapping[str, Any],
+                    src_streams: Sequence['Stream'],
+                    sensors: Mapping[str, Any]) -> _S: ...   # pragma: nocover
+
+
+class CamHttpStream(Stream):
+    """A cam.http stream."""
```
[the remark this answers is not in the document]
bmerry: Done

bmerry:
@sratcliffe I've made a pass over all your comments - thanks for the fast turnaround! A few of them need a bit more discussion before I can resolve them.

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
sratcliffe: Looks like a good list - I may be tempted to put the _VOL definitions in there as well, unless there is a reason to keep them in generator.

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
sratcliffe: Just figured the source string would possibly be more useful, and it less code to produce it.

katsdpcontroller/product_config.py
sratcliffe: Fair enough - perhaps n_instances ? I just don't like n_servers by itself since it carries hints of a mapping to physical hardware which is not necessarily true. Although we are getting dangerously close to the shed with this one :)

katsdpcontroller/product_config.py
sratcliffe: Fine - I just felt that 'Sensor foo has status happy' doesn't convey our dismay at the status being unexpected. In fact, how about "...has unexpected status..."

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
sratcliffe: Thanks - that makes more sense now.

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
bmerry: It's a bit of a bikeshed, but my feeling is that source strings are rather long it's then tedious to match up against all the long strings in the catalogue (or even just to find the point at which the source string ends and the error message begins).

katsdpcontroller/generator.py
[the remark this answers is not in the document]
sratcliffe: Yip

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
bmerry: I consider the _VOL constants to be closely tied to the details of how containers are run, rather than being tunable knobs like the others. I'd also rather not have a dependency on `scheduler`.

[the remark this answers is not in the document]
bmerry: I've moved the items from that list into a defaults.py. I'm not totally convinced that KATCBFSIM_SPECTRA_PER_HEAP truly belongs there, because it's describing a property of an external system rather than being an independently twiddlable knob, but that's getting into bikeshed territory.

katsdpcontroller/product_config.py
bmerry: I see your point, but the term "server" is also used in the command-line interfaces to katsdpingest, katcbfsim, and katsdpcal, so I'd rather stay consistent with that (and trying to change all of those to use "instances" and dealing with the compatibility issues doesn't seem worth the trouble).

katsdpcontroller/product_config.py
bmerry: I'm happier with that. Done.

katsdpcontroller/generator.py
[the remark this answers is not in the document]
bmerry: Done.

katsdpcontroller/product_config.py
sratcliffe: OK

katsdpcontroller/test/test_product_config.py
```diff
@@ -1,73 +1,1012 @@
...
+            product_config._normalise_output_channels(100, (-1, 1))
+
+    def test_align(self) -> None:
+        assert_equal(product_config._normalise_output_channels(1000, (299, 301), 100), (200, 400))
+        assert_equal(product_config._normalise_output_channels(1000, (200, 400), 100), (200, 400))
+
+    def test_misalign(self) -> None:
+        with assert_raises_regex(
+                ValueError,
+                r'n_chans \(789\) is not a multiple of required alignment \(100\)'):
```
sratcliffe, line 148: Just a question: Testing against an exact human readable error string always seems quite brittle to me. Wouldn't it be better to organise the testing so you cover the unrelated ValueError cases in another way, thus giving you confidence that the ValueError here is the intended one. Or better still, for calls that you can't guarantee unambiguous exceptions, rather have subclasses exceptions.

katsdpcontroller/test/test_product_config.py
```diff
@@ -1,73 +1,1012 @@
...
+        assert_equal(options.service_overrides, {})
+
+
+class TestSimulation:
+    """Test :class:`~.Simulation`."""
+
+    def test_from_config(self) -> None:
+        config = {
+            'sources': [
+                'PKS 1934-63, radec, 19:39:25.03, -63:42:45.7, (200.0 12000.0 -11.11 7.777 -1.231)',
```
sratcliffe, line 205: Nitpick: You have antennas defined at the top of the tests, why not sources as well ?

katsdpcontroller/test/test_product_config.py
```diff
@@ -1,73 +1,1012 @@
...
+
+
+class TestAntennaChannelisedVoltageStream:
+    """Test :class:`~.AntennaChannelisedVoltageStream`."""
+
+    def test_from_config(self) -> None:
+        config = {
+            'type': 'cbf.antenna_channelised_voltage',
+            'url': 'spead://239.0.0.0+7:7148',
+            'antennas': ['m000', 'm001'],
```
sratcliffe, line 250: Any point in adding some naming diversity here (e.g. s0002) to help prevent name interpretation that might sneak in ?

katsdpcontroller/test/test_product_config.py
```diff
@@ -1,73 +1,1012 @@
...
+        del self.config['excise']
+        vis = VisStream.from_config(Options(), 'sdp_l0', self.config, [self.bcp], {})
+        assert_equal(vis.output_channels, (0, 32768))
+        assert_equal(vis.excise, True)
+
+    def test_bad_continuum_factor(self) -> None:
+        self.config['continuum_factor'] = 3
+        with assert_raises_regex(
+                ValueError,
+                r'n_chans \(32768\) is not a multiple of required alignment \(12\)'):
```
sratcliffe, line 591: (Again with the string testing I know) - how about at least substituting in n_chans and cont_factor * n_servers for the 32768 and 12 values ?

katsdpcontroller/test/test_product_config.py
```diff
@@ -1,73 +1,1012 @@
...
+        assert_equal(bf.name, 'beamformer')
+        assert_equal(bf.antenna_channelised_voltage.name, 'narrow1_acv')
+        assert_equal(bf.tied_array_channelised_voltage, self.tacv)
+        assert_equal(bf.n_chans, 32768)
+
+    def test_mismatched_sources(self) -> None:
+        acv = make_antenna_channelised_voltage()
+        self.tacv[1] = make_tied_array_channelised_voltage(
+            acv, 'beam_0y', yarl.URL('spead://239.10.1.0+255:7148'))
+        with assert_raises_regex(ValueError,
```
sratcliffe, line 643: This test checks that the acv's are not the same object. Shouldn't it actually check that the actual config of the ACV is different ?

katsdpcontroller/test/test_product_config.py
```diff
@@ -1,73 +1,1012 @@
...
+        }
+
+    def test_from_config(self) -> None:
+        spec = SpectralImageStream.from_config(
+            Options(), 'spectral_image', self.config, [self.flags, self.continuum_image], {}
+        )
+        # Check that it gets copied
+        self.config['parameters'].clear()
+        assert_equal(spec.name, 'spectral_image')
+        assert_equal(spec.output_channels, (64, 100))
```
sratcliffe, line 904: Wouldn't it be more robust to compare to the config dict - assert_equal(spec.output_channels, self.config['output_channels']) ?

katsdpcontroller/test/test_product_config.py
bmerry: I decided to put the full string in mostly because I have a bad habit of returning error strings that contain curly braces because I forgot to call `.format` on them (or make them f-strings, in 3.6+). This ensures not just that I'm covering the right piece of code, but that the error string is actually useful to a human.

katsdpcontroller/test/test_product_config.py
bmerry: The antennas are used in multiple tests, while these sources are only used here. If they get used elsewhere I'll make them global.

katsdpcontroller/test/test_product_config.py
bmerry: I'm pretty confident that I haven't done any such thing in the master controller code.

katsdpcontroller/test/test_product_config.py
bmerry: Done

katsdpcontroller/test/test_product_config.py
bmerry: Ok, I've made an ACV that is configured with different antennas.

katsdpcontroller/test/test_product_config.py
bmerry: Done

katsdpcontroller/test/test_product_config.py
sratcliffe: My point is more about getting into the habit of using different antenna names, just because it emphasises the fact that we have different ones. I would never cast such an aspersion on your person, but lesser mortals have been known to fall into this trap....

katsdpcontroller/test/test_product_config.py
sratcliffe: Fair enough

katsdpcontroller/test/test_product_config.py
sratcliffe: You still haven't quite sold me on this. In general I would not really expect changing the log message of an exception slightly to result in the unit test failing. Not sure there is a better solution though... Maybe @ludwigschwardt has a view ?

ludwigschwardt: The only other option to check exceptions that I can think of is to return some of these numbers as exception arguments, but that might be overkill again.

bmerry: That still wouldn't guarantee that the exception *message* is properly formatted.

katsdpcontroller/test/test_product_config.py
bmerry: Ok, I've sprinkled some alternative names around the tests.

katsdpcontroller/test/test_product_config.py
sratcliffe: We could have a messages library which at least provides a single source for these text strings, but by that stage we are multiple coats deep on the bikeshed. I wave the white flag of defeat :)

PR #488, opened 2020-07-15
Stop publishing katcp sensors for batch tasks

bmerry:
There were .version, .state and .mesos-state sensors being published,
but this has caused problems for CAM which doesn't cope well with a
large number of sensors combined with a high frequency of
interface-changed notifications.

It also adds a missing interface-changed notification when a task with
these sensors starts up.

This should mitigate MKAIV-1593.


bmerry:
@ajoubertza FYI so you know this is in the pipe. It'll probably need some real-world testing to ensure that I've found all the appropriate `interface-changed` calls.

PR #489, opened 2020-07-16
Double memory for the product controller

bmerry:
We've had some OOM issues (SPR1-614). It shouldn't be using anywhere
near this much memory, but this is a short-term mitigation.


bmerry:
This is one of the rare changes that will require a master controller restart.

bmerry:
> Thanks - at some stage we should scale this with channel count as well (just in case imaging is triggered).

This is just a stop-gap - I need to figure out what's eating all the memory, given that there shouldn't be more than kilobytes per task.

PR #494, opened 2020-07-21
Optimise getting sensors from katportal

bmerry:
katportal on MeerKAT is very slow (takes 1.5s+ to get each sensor). I'm
guessng it lacks a dictionary lookup and iterates over every single
sensor. Speed it up by munging all the sensors to query into a regex and
querying with that, which at least only makes one pass over the sensor
list on the backend.

This should fix SPR1-621.

### DIFF
--- a/katsdpcontroller/product_config.py
+++ b/katsdpcontroller/product_config.py
@@ -5,6 +5,7 @@ import logging
 import itertools
 import copy
 import math
+import re
 from abc import ABC, abstractmethod
 from distutils.version import StrictVersion
 from typing import (
@@ -1224,9 +1225,9 @@ class Configuration:
                 cam_http = CamHttpStream.from_config(options, name, stream_config, [], {})
                 break
 
-        # Extract sensor values. This is pulled into a separate block rather
-        # than done as we construct each stream so that one day it can be
-        # optimised to make a single sensor_values call for all sensors.
+        # Extract sensor values. This is pulled into a separate block that
+        # queries all sensors at once rather than done as we construct each
+        # stream because katportal has a very high overhead per query.
         sensors: Dict[str, Dict[str, Any]] = {name: {} for name in stream_configs}
         if cam_http:
             client = katportalclient.KATPortalClient(str(cam_http.url), None)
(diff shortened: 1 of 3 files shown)


bmerry:
@ajoubertza just so that you're aware that the performance of querying sensors one-at-a-time is abysmal.

ajoubertza:
@bmerry Yes, it very slow.  We have a [ticket](https://skaafrica.atlassian.net/browse/MT-1078) for improvements on our side, but unfortunately it isn't high priority.

Your regex approach is good.  FBFUSE had to do something similar.

bmerry:
@ajoubertza thanks, I've added a reference to that ticket in a comment. I'm not 100% sure that ticket applies because we were using `sensor_value`, not `sensor_values`, and the former takes a name rather than a regex, but maybe you're turning that into a regex on the backend.

ajoubertza:
@bmerry both `sensor_value` and `sensor_values` use the same endpoint on portal.  Yes, we add the regex for `sensor_value` in the background.

katsdpcontroller/product_config.py
```diff
@@ -1224,9 +1225,9 @@ async def from_config(cls, config: Mapping[str, Any]) -> 'Configuration':
                 cam_http = CamHttpStream.from_config(options, name, stream_config, [], {})
                 break
 
-        # Extract sensor values. This is pulled into a separate block rather
-        # than done as we construct each stream so that one day it can be
-        # optimised to make a single sensor_values call for all sensors.
+        # Extract sensor values. This is pulled into a separate block that
+        # queries all sensors at once rather than done as we construct each
```
sratcliffe, line 1229: pedant: "... at once, rather than..."

katsdpcontroller/product_config.py
```diff
@@ -1238,15 +1240,30 @@ async def from_config(cls, config: Mapping[str, Any]) -> 'Configuration':
...
             for name, stream_config in stream_configs.items():
                 stream_cls = STREAM_CLASSES[stream_config['type']]
                 for sensor in stream_cls._class_sensors:
                     instrument = stream_config['instrument_dev_name']
                     full_name = sensor.full_name(components, name, instrument)
-                    try:
-                        sample = await client.sensor_value(full_name)
-                    except Exception as exc:
-                        raise SensorFailure(f'Could not get value for {full_name}: {exc}') from exc
+                    if full_name not in samples:
```
sratcliffe, line 1264: I can't really see the logic of this second loop. Can't we just check for full_names == samples to check for the case of a missing sensor, and then just loop through samples to check the status of each ?

bmerry: I'm not following - how would we know the values of `name`, `sensor.name` and `sensor.type` in that case?

sratcliffe: Why do we need them ? We have already iterated through stream_configs to populate full_names, so we have the names that should appear in samples, so checking to see if a requested sensor doesn't exist is easy (i.e. we don't need to loop through stream_configs again). And then only other thing this block does is check the state of each sample...

bmerry: Expand the unmodified lines in the diff - there are unmodified lines further down the block that need them.

sratcliffe: Fair enough, I yield, although I do think you could still do type_checking and initial assignment without the duplicated code block by just caching the found sensors from the first loop.

bmerry:
> Reluctant approval :)

I wasn't totally happy with the duplicated loop myself - I'm happy to improve it if it can be done without adding complexity somewhere else. But I'm not following your solution, so it would help if you could put it in the form of actual code.

sratcliffe:
Yip - I can only really make it two lines shorter and it doesn't gain much :) Let's leave it as is...

PR #495, opened 2020-07-21
Fix postprocessing failing if the queue is blocked by an earlier one

bmerry:
If a second postprocessing task is started and there are tasks from the
first that haven't cleared the batch queue, then it would fail because
it would only wait the default resources_timeout (11s) to launch the
pseudo-task that wraps telstate.

Fixes SPR1-628.

### DIFF
--- a/katsdpcontroller/scheduler.py
+++ b/katsdpcontroller/scheduler.py
@@ -1221,9 +1221,10 @@ class GroupInsufficientInterfaceResourcesError(InsufficientResourcesError):
 
 class QueueBusyError(InsufficientResourcesError):
     """The launch group did not reach the front of the queue before its timeout expired."""
-    def __init__(self):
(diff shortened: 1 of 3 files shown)


katsdpcontroller/scheduler.py
```diff
@@ -1219,6 +1219,14 @@ def __str__(self):
                 "to launch all tasks ({0.needed} > {0.available})".format(self))
 
 
+class QueueBusyError(InsufficientResourcesError):
+    """The launch group did not reach the front of the queue before its timeout expired."""
+    def __init__(self):
+        super().__init__(
+            "The launch group did not reach the front of the queue before its timeout expired"
```
sratcliffe, line 1226: Please include the timeout itself just to aid debugging.

bmerry: Done.

PR #499, opened 2020-08-24
Skip spectral imaging tasks that are completely band masked

bmerry:
Only the RFI mask was being applied to compute the channel mask; now the
band mask is used as well.

This partially addresses SPR1-649. Support is also needed in the imager
to skip individual channels when the 128-channel batch lies partially
inside and partially outside the masked region.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -19,6 +19,7 @@ import katsdptelstate.aio
 from katsdptelstate.endpoint import Endpoint
 import katsdpmodels.fetch.aiohttp
 from katsdpmodels.rfi_mask import RFIMask
+from katsdpmodels.band_mask import BandMask, SpectralWindow
 
 from . import scheduler, product_config, defaults
 from .tasks import (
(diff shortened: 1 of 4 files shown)


katsdpcontroller/product_config.py
```diff
@@ -1153,6 +1153,10 @@ def n_chans(self) -> int:
     def flags(self) -> FlagsStream:
         return cast(FlagsStream, self.src_streams[0])
 
+    @property
```
sratcliffe, line 1156: ImageStream already has this property, why do we need to define it again ?

bmerry: Thanks, well spotted! It looks like I'm also uselessly redefining the flags property. I'll get rid of both.

sratcliffe (approved):
Looks good

katsdpcontroller/generator.py
```diff
@@ -1578,14 +1579,22 @@ async def _make_spectral_imager(g: networkx.MultiDiGraph,
...
 
     async with katsdpmodels.fetch.aiohttp.TelescopeStateFetcher(telstate) as fetcher:
         rfi_mask = await fetcher.get('model_rfi_mask_fixed', RFIMask)
         max_baseline = rfi_mask.max_baseline_length(channel_freqs)
         channel_mask = max_baseline > 0 * u.m
+        acv = stream.vis.baseline_correlation_products.antenna_channelised_voltage
+        telstate_acv = telstate.view(acv.name)
+        band_mask = await fetcher.get('model_band_mask_fixed', BandMask, telstate=telstate_acv)
+        channel_mask |= band_mask.is_masked(
+            SpectralWindow(spw.bandwidth * u.Hz, spw.centre_freq * u.Hz),
```
bmerry, line 1595: I've realised that this is wrong - the bandwidth of the data that reaches katdal might not be the full band. I'll make an update that instead gets the data from the CAM sensors.

PR #500, opened 2020-08-24
Fix calculation of band mask for spectral imaging

bmerry:
It was using the spectral window reported by katdal, but if ingest is
configured to consume only part of the band that will be incorrect. It
must use the full band (from the AntennaChannelisedVoltage stream).


sratcliffe (approved):
Nice catch

PR #504, opened 2020-11-06
Correctly handle unavailability information in offers

bmerry:
Fixes SPR1-700.


sratcliffe (approved):
Seems good - surprised we even get offers with no specified end time...

bmerry:
No duration means "forever". We only started using that when I rewrote mesosctl.

PR #505, opened 2020-11-06
Fix service overrides

bmerry:
They were broken by the product_config refactoring for API v3, which introduced ServiceOverride in place of a dictionary.


sratcliffe (approved):
Looks good

PR #506, opened 2020-11-30
Make Mesos responsible for timing out batch tasks

bmerry:
This simplifies the code slightly, and ensures that tasks will be killed
off even if there is a network partition or the product controller gets
wedged.

Some extra information is added to TaskError messages to report the
reason (which for these timeouts is REASON_MAX_COMPLETION_TIME_REACHED)
so that it will be clear why the task failed.

Closes SPR1-488.


richarms (approved):
Takes advantage of Mesos framework support (as of v. 1.6.0 May 11, 2018) for task timeout handling.

PR #507, opened 2020-11-30
More compact formatting of durations on dashboard

bmerry:
Instead the rather clunky "6 hours, 32 minutes and 7.73 seconds", show
6:32:08. This removes the need for the humanfriendly package, and causes
less visual layout disruption when it is ticking because it has a mostly
fixed width (except when the number of digits in the hours increases).


richarms (approved):
This seems reasonable.

PR #508, opened 2020-12-01
Switch multicast to use 239.192.0.0/18 address space

bmerry:
This is RFC2365-compliant, and should play nice with the CBF network
which only supports the 239.0.0.0/8 range. I switched from a /16 to a
/18 because we don't need a lot of addresses and the RFC2365
organisation-local scope is not very big (/14).

Relates to SPR1-706.

### DIFF
--- a/katsdpcontroller/master_controller.py
+++ b/katsdpcontroller/master_controller.py
@@ -1568,7 +1568,7 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
                         metavar='FILE',
                         help='configuration for connecting services to S3 '
                              '(loaded on each configure)')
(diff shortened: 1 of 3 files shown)


bmerry:
@sratcliffe FYI since you were involved in the discussion leading to this.

richarms:
16,382 seems like enough..

bmerry:
Thanks @richarms. Just a note on our workflow: in most cases the reviewer will approve a PR and the submitter will do the actual merge. This gives the submitter a chance to control the timing of the merge (e.g. they might have a closely-related PR on another package with a different reviewer and they need to be merged together). In this case I would have just merged myself so no harm done.

PR #511, opened 2020-12-15
Switch to total intensity imaging only

sratcliffe:


bmerry:
Is the intention to make this a permanent change (i.e. merge it) or just to use it for the holiday?

sratcliffe:
My thinking is to make it permanent until such time as CAM properly allows users to en/disable the pipelines and specify some user configuration. Feedback is that almost all users are happy with just I, and with the constraints on the archive and link it is probably better to stick with this, and have full stokes as the exception.

bmerry:
Ok. I'm approving, pending testing.

bmerry (approved):

PR #512, opened 2021-01-19
Preparations for moving to Ubuntu 20.04 base image

bmerry:
- Install the distro's haproxy instead of building from source.
- Bump http-parser, kazoo to versions that works on Python 3.8.


bmerry:
Marked as draft because it depends on katsdpdockerbase being updated (at which point the Dockerfile here needs to be updated to point at the master branch of that again).

richarms (approved):

PR #516, opened 2021-02-22
Adjust the resource requirements for spectral imager

bmerry:
The memory and GPU requirements have been reduced (based on examining
resource usage from a 7 hour observation) to allow 8 processes per
server rather than 4. A smaller-scale test (2 processes in place of 1,
restricted to 6 cores) suggests each process only becomes slightly
slower, and hence overall throughput should be significantly higher.


bmerry:
@lmzxqtom sending this your way since it will put more pressure on nodes that also happen to run on Ceph. In practice we'll just have to test it and see how it works.

lmzxqtom:
@bmerry sounds like an interesting test :)

The current Ceph OSD memory target on the imgr machines is set to 3 GB.  I could set a more aggressive memory target of 2GB if it looks like the memory is a contention.

lmzxqtom (approved):

PR #517, opened 2021-03-07
Some requirements cleanup

bmerry:
- Explicitly list numpy in setup.py (it's used directly)
- Add a pyproject.toml intead of listing katversion in requirements.txt
- Sort requirements in setup.py alphabetically
- List extras on git dependencies in requirements.txt
- Add -c option to requirements.txt to set defaults
- Remove recursive requirements that have default versions
- Unpin asynctest, since it is pinned in the base requirements

See SPR1-833.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ nbformat==4.4.0            # via plotly
 netifaces
 networkx==2.0
 numpy
-plotly==4.0.0              # via plotly
+plotly==4.0.0              # via dash
 prometheus_async==18.1.0
 prometheus_client==0.2.0
 pydotplus==2.0.2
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,91 +1,53 @@
...
 monotonic==1.5             # via prometheus_async
-msgpack                    # via katsdptelstate
-multidict                  # via aiohttp, yarl
 nbformat==4.4.0            # via plotly
-numba                      # via katdal
 netifaces
 networkx==2.0
 numpy
-omnijson                   # via katportalclient
 plotly==4.0.0              # via plotly
```
ludwigschwardt, line 36: Plotly is not listed as a direct dependency, and it arrives "via plotly"... 🙂

ludwigschwardt (commented):
Just the question around plotly...

requirements.txt
bmerry: Heh, requirement-ception :-) Fixed the comment.

ludwigschwardt (approved):

richarms (approved):

PR #520, opened 2021-03-30
Minor changes to the sandbox doc

richarms:
One or two minor changes to the sandbox setup documentation
prepare.sh now creates logs directory


sandbox/sdp_image_tag
```diff
@@ -0,0 +1,2 @@
+latest
+
```
bmerry, line 2: I don't think it'll make a difference, but delete the blank line.

bmerry (changes requested):
Looks good, just one change for my whitespace OCD.

bmerry (approved):

PR #521, opened 2021-04-07
remove lingering references to interface_mode argument

richarms:
Since the interface_mode argument no longer exists,  remove references to it


bmerry (changes requested):
Looks like some nice cleanup. Can you also move the argument itself from `add_shared_options` into `parse_args` in `master_controller.py`.

bmerry (approved):
Great, thanks.

PR #522, opened 2021-04-13
Fix product controller after interface mode cleanup

bmerry:
7b9c0af removed the code that passed interface_mode to DeviceServer, but
it's a required argument. Add it back.

Also remove sandbox/sdp_image_tag, because it's local configuration and
if it's in version control, it would be very easy for someone to test on
a branch, accidentally check in the file, and then someone else will
pull down the change and end up working with a different tag to the one
they thought they were. Add it to .gitignore instead.


ludwigschwardt (approved):
Looks good!

bmerry:
@richarms FYI. I got Ludwig to review because you're on holiday. I'm not sure if you'll want to add setup instructions to create the sdp_image_tag the first time.

PR #523, opened 2021-04-15
Limit spectral imaging to one per GPU again

bmerry:
This should reduce the chance of starving Ceph of resources e.g. as per
SPR1-904. I've put the memory limit to 12GB; this is a compromise
between the original 15GB (which when used could starve Ceph) and the
more recent 8GB (which was fairly aggressive to ensure that we could fit
8 imagers per box).

This partially undoes #516.


bmerry:
@lmzxqtom assigning to you just because you reviewed the original #516, but let me know if I should get someone else to do it.

lmzxqtom (approved):
Thanks Bruce. A pity to reduce this functionality.

PR #525, opened 2021-06-03
bump dash >=1.20.0

richarms:
Neglected to update requirements.txt to update version of dash (>=1.20.0) during https://skaafrica.atlassian.net/browse/SPR1-868

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,12 +8,11 @@ aioredis
 aiozk==0.14.0
 async-timeout
 click==7.0                 # via Flask
-dash==1.0.2
-dash-core-components==1.0.0
-dash-html-components==1.0.0
-dash-table==4.0.2
-dash-dangerously-set-inner-html==0.0.2
-dash-renderer==1.0.0       # via dash
+dash==1.20.0
+dash_core_components==1.16.0
+dash_dangerously_set_inner_html==0.0.2
+dash_html_components==1.1.3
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -8,12 +8,11 @@ aioredis
 aiozk==0.14.0
 async-timeout
 click==7.0                 # via Flask
-dash==1.0.2
-dash-core-components==1.0.0
-dash-html-components==1.0.0
-dash-table==4.0.2
-dash-dangerously-set-inner-html==0.0.2
-dash-renderer==1.0.0       # via dash
```
bmerry, line 16: Interesting that this vanishes - did it get absorbed into dash itself?

setup.py
```diff
@@ -30,7 +30,7 @@
         'aioredis',
         'aiozk',
         'async_timeout',
-        'dash',
+        'dash>=1.18.*',
```
bmerry, line 33: Not sure that `*` makes logical sense with `>=`, but presumably pip is happy enough or this wouldn't have built). I would have said `dash>=1.18`.

bmerry (approved):

requirements.txt
richarms: yes, it's obsolete -- merged into the main dash repo

setup.py
richarms: noted

PR #526, opened 2021-06-11
SPR1-1032: put links to primary beams into telstate

bmerry:
For the sandbox environment, I'm copying the real primary beams from sdpmodels rather than trying to make up a model.

### DIFF
--- a/katsdpcontroller/test/test_product_controller.py
+++ b/katsdpcontroller/test/test_product_controller.py
@@ -34,6 +34,7 @@ import katpoint
 import katdal
 import katsdpmodels.band_mask
 import katsdpmodels.rfi_mask
+import katsdpmodels.primary_beam
 import yarl
 
 from ..controller import device_server_sockname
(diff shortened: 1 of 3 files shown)


katsdpcontroller/test/test_product_controller.py
```diff
@@ -346,6 +346,24 @@ def _setup_band_mask_model(self) -> None:
             '/models/band_mask/fixed/test.h5'
         )
 
+    def _setup_primary_beam_model(self) -> None:
+        # Model is not used for anything, so do a minimal amount to get it working.
+        model = katsdpmodels.primary_beam.PrimaryBeamAperturePlane(
+            -1 * u.m, -1 * u.m,
+            1 * u.m, 1 * u.m,
+            [1, 2] * u.GHz,
+            np.zeros((2, 2, 2, 8, 8), np.complex64),
```
ludwigschwardt, line 355: Why 8x8? I only see -1 m start and 1 m step. Where does the 8 happen?

ludwigschwardt (approved):

katsdpcontroller/test/test_product_controller.py
bmerry: It's the number of steps. It happens right here - it doesn't need to be repeated elsewhere.

ludwigschwardt: Ah, right, gotcha. It _might_ have been clearer if start = -3, step = 1 and number of steps = 7, suggesting a symmetrical grid on the aperture plane, but I don't think it's worth changing.

bmerry: Mattieu's models aren't symmetric either :-) Although less asymmetric (just one pixel) than this ended up being. As you say, not worth changing.

PR #527, opened 2021-06-23
NGC-189: Add new stream types for next-gen correlator

bmerry:
This just accepts them in a configuration dictionary, but doesn't do anything with them, and any attempt to hook them up to SDP components will almost certainly fail as the appropriate sensors won't exist.

The specification for the schema updates is [here](https://docs.google.com/document/d/1QBlph42ExntZaKf1iRB5SaGsMwZEvrO4cFZ5ot_E7eg/edit?usp=sharing).

### DIFF
--- a/katsdpcontroller/product_config.py
+++ b/katsdpcontroller/product_config.py
@@ -441,11 +441,11 @@ class NgcAntennaChannelisedVoltageStream(AntennaChannelisedVoltageStreamBase):
             if src.adc_sample_rate != first.adc_sample_rate:
                 raise ValueError(
                     'Inconsistent ADC sample rates '
-                    f'(both {first.adc_sample_rate} and {src.adc_sample_rate}')
+                    f'(both {first.adc_sample_rate} and {src.adc_sample_rate})')
             if src.centre_frequency != first.centre_frequency:
                 raise ValueError(
                     'Inconsistent centre frequencies '
-                    f'(both {first.centre_frequency} and {src.centre_frequency}')
+                    f'(both {first.centre_frequency} and {src.centre_frequency})')
         super().__init__(
             name, src_streams,
             antennas=antenna_names,
@@ -635,6 +635,15 @@ class BaselineCorrelationProductsStreamBase(CbfPerChannelStream):
         """Size of frame in bytes"""
         return self.n_vis * 2 * self.bits_per_sample // 8
 
+    @staticmethod
+    def round_int_time(int_time: float,
+                       acv: AntennaChannelisedVoltageStreamBase,
+                       spectra_per_heap: int) -> float:
+        """Round an integration time to a positive integer number of heaps."""
+        heap_time = acv.n_chans / acv.bandwidth * spectra_per_heap
+        acc_heaps = max(1, round(int_time / heap_time))
+        return acc_heaps * heap_time
+
 
 class BaselineCorrelationProductsStream(CbfStream, BaselineCorrelationProductsStreamBase):
     """Real baseline-correlation-products stream (old correlator)."""
@@ -697,13 +706,13 @@ class NgcBaselineCorrelationProductsStream(BaselineCorrelationProductsStreamBase
 
     def __init__(self, name: str, src_streams: Sequence[Stream], *,
                  int_time: float) -> None:
-        # TODO: how should the number of engines be computed?
         acv = src_streams[0]
         assert isinstance(acv, AntennaChannelisedVoltageStreamBase)
         n_ants = len(acv.antennas)
-        n_engines = 8
+        n_engines = 8   # TODO: need to compute this from number of antennas
         if acv.n_chans % n_engines != 0:
             raise ValueError('Number of channels is not a multiple of the number of X-engines')
+        int_time = self.round_int_time(int_time, acv, defaults.NGC_SPECTRA_PER_HEAP)
         super().__init__(
             name, src_streams,
             int_time=int_time,
@@ -750,9 +759,7 @@ class SimBaselineCorrelationProductsStream(BaselineCorrelationProductsStreamBase
         n_antennas = len(acv.antennas)
         # Round the int_time the same way katcbfsim does, so that we have
         # an accurate value.
-        heap_time = acv.n_chans / acv.bandwidth * defaults.KATCBFSIM_SPECTRA_PER_HEAP
-        acc_heaps = max(1, round(int_time / heap_time))
-        int_time = acc_heaps * heap_time
+        int_time = self.round_int_time(int_time, acv, defaults.KATCBFSIM_SPECTRA_PER_HEAP)
         super().__init__(
             name, src_streams,
             int_time=int_time,
--- a/katsdpcontroller/product_config.py
+++ b/katsdpcontroller/product_config.py
@@ -59,6 +59,18 @@ def _url_n_endpoints(url: Union[str, yarl.URL]) -> int:
     return len(endpoint_list_parser(None)(url.host))
 
 
+def _make_antenna(description: str) -> katpoint.Antenna:
+    """Create an antenna object from a description.
+
+    The :class:`katpoint.Antenna` constructor can throw assorted errors, so this
+    helper method always raises :exc:`ValueError` if the description is invalid.
+    """
+    try:
+        return katpoint.Antenna(description)
+    except Exception as exc:
+        raise ValueError(f'Invalid antenna description {description!r}: {exc}') from exc
+
+
 class SensorFailure(RuntimeError):
     """Failed to obtain a sensor value from katportal"""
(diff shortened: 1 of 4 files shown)


katsdpcontroller/product_config.py
```diff
@@ -59,6 +59,18 @@ def _url_n_endpoints(url: Union[str, yarl.URL]) -> int:
...
 
+def _make_antenna(description: str) -> katpoint.Antenna:
+    """Create an antenna object from a description.
+
+    The :class:`katpoint.Antenna` constructor can throw assorted errors, so this
+    helper method always raises :exc:`ValueError` if the description is invalid.
+    """
+    try:
+        return katpoint.Antenna(description)
+    except Exception as exc:
```
james-smith-za, line 70: Just catching an `Exception` seems risky to me. Doesn't katpoint already raise a `ValueError` if the `description` string is bad?

katsdpcontroller/product_config.py
```diff
@@ -296,6 +308,34 @@ class CbfStream:
...
+        self.band = band
+        self.antenna = antenna
+
+    @classmethod
+    def from_config(cls,
+                    options: Options,
+                    name: str,
+                    config: Mapping[str, Any],
+                    src_streams: Sequence['Stream'],
+                    sensors: Mapping[str, Any]) -> 'SimDigRawAntennaVoltageStream':
```
james-smith-za, line 331: `sensors` argument doesn't appear to be used?

katsdpcontroller/product_config.py
```diff
@@ -374,6 +414,62 @@ def from_config(cls,
...
+
+    It currently only supports wideband.
+    """
+
+    stream_type: ClassVar[str] = 'ngc.antenna_channelised_voltage'
+    _valid_src_types: ClassVar[_ValidTypes] = {'sim.dig.raw_antenna_voltage'}
+
+    def __init__(self, name: str, src_streams: Sequence[Stream], *,
+                 n_chans: int) -> None:
+        if n_chans < 1 or (n_chans & (n_chans - 1)) != 0:
```
james-smith-za, line 428: Nice trick! I'm going to add that to my mental toolbox to check for a power of 2.

katsdpcontroller/product_config.py
```diff
@@ -374,6 +414,62 @@ def from_config(cls,
...
+            raise ValueError('n_chans is not a power of 2')
+        if len(src_streams) % 2 != 0:
+            raise ValueError('src_streams does not have an even number of elements')
+        first = src_streams[0]
+        assert isinstance(first, SimDigRawAntennaVoltageStream)
```
james-smith-za, line 433: I'm ... wary of `assert` statements anywhere except in unit tests.* AFAIK if you run Python with the `-O` option then the statement just gets ignored.

* Although you could argue that sprinkling them in your code is useful for testing, knowing that they will have zero impact on the actual production code.

katsdpcontroller/product_config.py
```diff
@@ -374,6 +414,62 @@ def from_config(cls,
...
+            n_samples_between_spectra=2 * n_chans
+        )
+
+    @classmethod
+    def from_config(cls,
+                    options: Options,
+                    name: str,
+                    config: Mapping[str, Any],
+                    src_streams: Sequence['Stream'],
+                    sensors: Mapping[str, Any]) -> 'NgcAntennaChannelisedVoltageStream':
```
james-smith-za, line 466: `sensors` not used again. Is this a pattern or a convention that the argument list needs to have this in it?

katsdpcontroller/test/test_product_config.py
```diff
@@ -241,6 +245,35 @@ def test_from_config(self) -> None:
...
+            'band': 'l',
+            'antenna': _M000.description
+        }
+
+    def test_from_config(self) -> None:
+        dig = SimDigRawAntennaVoltageStream.from_config(
+            Options(), 'm000h', self.config, [], {}
+        )
+        assert_equal(dig.adc_sample_rate, 1712000000.0)
+        assert_equal(dig.centre_frequency, 1284000000.0)
```
james-smith-za, line 265: Perhaps my understanding of the unit-testing philosophy is wrong here.
Would it not be better to use references to `self.config.adc_sample_rate` for instance? Hardcoding the number seems prone to problems if you change one but forget to change the other.

james-smith-za (commented):
I am not familiar with this module at all. I gave it as careful a look as I know how. Didn't spot any major issues.

katsdpcontroller/product_config.py
bmerry: That's the problem: there are assorted errors that can occur (I don't recall exactly what, since this is code I wrote long ago and refactored here).

katsdpcontroller/product_config.py
bmerry: Correct. All the stream classes define this function, and there is generic code that looks up the class to build at run-time and calls its `from_config` method. So they all need to have the same signature.

katsdpcontroller/product_config.py
bmerry: In general (and in many languages), `assert` statements in non-test code are for checking for conditions that indicate bugs (whereas exceptions generally indicate that the code throwing the exception has encountered an exceptional condition but is not necessarily buggy). They're turned off in optimised code because you wouldn't ship buggy code, right? :-) And then they help in debugging by exploding as soon as the bug is detected, rather than 10 steps later in some obscure way.

In this specific case, it's related to mypy. It recognises assert statements like this and after this line assumes that `first` is an instance of `SimDigRawAntennaVoltageStream` and can type-check it appropriately.

katsdpcontroller/test/test_product_config.py
bmerry: You're right, I am violating the don't-repeat-yourself principle. I've pushed some changes. I tend to prefer putting explicit values in rather than formulae because I've been burned before by having the formula be incorrect, whereas if I evaluate the formula by hand I can look at the result and say "that's not right" (once at an old job it was due to an `int` being cast to a C 32-bit `float` and losing precision in the process, which you couldn't immediately see by eye-balling the code). But in these cases the value I need is right there in the dictionary, no formula required.

katsdpcontroller/product_config.py
```diff
@@ -318,7 +358,7 @@ def __init__(self, name: str, src_streams: Sequence[Stream], *,
 
 
 class AntennaChannelisedVoltageStream(CbfStream, AntennaChannelisedVoltageStreamBase):
-    """Real antenna-channelised-voltage stream."""
+    """Real antenna-channelised-voltage stream (old correlator)."""
```
ajoubertza, line 361: I'm wondering about the naming.  "Next-generation" and "old" make sense now, but what happens some time in the future when we bring out the next next-generation correlator?

bmerry: This is the topic of https://skaafrica.atlassian.net/browse/NGC-199 :-) I'll update once we have a resolution there.

ajoubertza: Perfect 🙂

katsdpcontroller/product_config.py
james-smith-za: I suppose catching and re-raising it again like this works then.
But you lose the original type, you raise it as a `ValueError`. You do preserve the message though I guess.

james-smith-za: Actually cancel that. I went to play around with what this does if you use `raise .. from .. ` and the behaviour is sane. You see both exceptions.

bmerry: Yes, losing the original type is the point. The katpoint interface is nasty because you get an exception of some arbitrary type depending on what it was trying to do (maybe you get an IndexError if you didn't pass enough fields or something), but I want the product_config constructors to predictably raise ValueError if they're given bad input.

katsdpcontroller/product_config.py
james-smith-za: Right. Python and its dynamic typing coming to create philosophical conundrums.

Your explanation makes sense. I'll allow it. I'll have to do some reading up.

katsdpcontroller/test/test_product_config.py
james-smith-za: Changes look good to me.

But you do also make a point regarding evaluating test cases "by hand" so to speak. I'd just want stuff to be reviewed very very carefully if we do that. I'm good at making silly errors in by-hand calculations.

bmerry: As long as it doesn't exactly match a mistake in the calculation coded in the implementation, the unit test will tell you :-)

james-smith-za (approved):

james-smith-za:
That's a good point.

PR #530, opened 2021-06-29
NGC-189: Add input_labels and command_line_extra options to schema

bmerry:
The former is to support assigning unique labels to digitiser outputs
even when repeating digitisers to make up extra antennas. The latter is
an escape hatch to allow for passing extra arguments directly to
processes without rebuilding the container.

### DIFF
--- a/katsdpcontroller/product_config.py
+++ b/katsdpcontroller/product_config.py
@@ -315,12 +315,14 @@ class SimDigRawAntennaVoltageStream(Stream):
                  adc_sample_rate: float,
                  centre_frequency: float,
                  band: str,
-                 antenna: katpoint.Antenna) -> None:
+                 antenna: katpoint.Antenna,
+                 command_line_extra: Iterable[str] = ()) -> None:
         super().__init__(name, [])
         self.adc_sample_rate = adc_sample_rate
         self.centre_frequency = centre_frequency
         self.band = band
         self.antenna = antenna
+        self.command_line_extra = list(command_line_extra)
 
     @classmethod
     def from_config(cls,
@@ -333,7 +335,8 @@ class SimDigRawAntennaVoltageStream(Stream):
                    adc_sample_rate=config['adc_sample_rate'],
                    centre_frequency=config['centre_frequency'],
                    band=config['band'],
-                   antenna=_make_antenna(config['antenna']))
+                   antenna=_make_antenna(config['antenna']),
+                   command_line_extra=config.get('command_line_extra', []))
 
 
 class AntennaChannelisedVoltageStreamBase(Stream):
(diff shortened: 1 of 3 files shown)


katsdpcontroller/product_config.py
```diff
@@ -704,8 +723,10 @@ class NgcBaselineCorrelationProductsStream(BaselineCorrelationProductsStreamBase
     stream_type: ClassVar[str] = 'ngc.baseline_correlation_products'
     _valid_src_types: ClassVar[_ValidTypes] = {'ngc.antenna_channelised_voltage'}
 
-    def __init__(self, name: str, src_streams: Sequence[Stream], *,
-                 int_time: float) -> None:
+    def __init__(self, name: str,
+                 src_streams: Sequence[Stream], *,
+                 int_time: float,
```
james-smith-za, line 728: `int_time` was confusing to me for a bit. Why is it called `int` time if it's a `float`? I take it you mean integration time?

katsdpcontroller/test/test_product_config.py
```diff
@@ -386,6 +396,32 @@ def test_centre_frequency_mismatch(self) -> None:
...
+    def test_input_labels(self) -> None:
+        self.config['input_labels'] = ['m900h', 'm900v', 'm901h', 'm901v']
+        acv = NgcAntennaChannelisedVoltageStream.from_config(
+            Options(), 'wide1_acv', self.config, self.src_streams, {}
+        )
+        assert_equal(acv.input_labels, self.config['input_labels'])
+
+    def test_bad_input_labels(self) -> None:
+        self.config['input_labels'] = ['m900h']
+        with assert_raises_regex(ValueError, 'input_labels has the wrong number of elements'):
```
james-smith-za, line 408: The change that I suggested will probably break this.

As an aside, why is this `assert_raises_regex()` function used instead of, say `@pytest.mark.xfail`?

james-smith-za (approved):
I made some comments but I don't consider anything to be a show-stopper.

After tomorrow's code walkthrough I will likely be in a better position to offer feedback in future.

katsdpcontroller/product_config.py
ludwigschwardt: :rofl: That's the official CBF term. 😛

katsdpcontroller/product_config.py
```diff
@@ -315,12 +315,14 @@ def __init__(self, name: str, src_streams: Sequence[Stream], *,
...
                  band: str,
-                 antenna: katpoint.Antenna) -> None:
+                 antenna: katpoint.Antenna,
+                 command_line_extra: Iterable[str] = ()) -> None:
         super().__init__(name, [])
         self.adc_sample_rate = adc_sample_rate
         self.centre_frequency = centre_frequency
         self.band = band
         self.antenna = antenna
+        self.command_line_extra = list(command_line_extra)
```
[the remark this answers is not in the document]
bmerry: An alternative would be to make the default value be an empty list, but linting tools tend to complain about mutable default values (even though it's safe in this case since it's not mutated). Or I could make `self.command_line_extra` be a tuple, but then I'd have to cast the list that comes out of the JSON parser to a tuple. Or I could make the default be None, but then I have to do something like `command_line_extra if command_line_extra else []` which I find less readable.

katsdpcontroller/product_config.py
```diff
@@ -423,12 +426,25 @@ class NgcAntennaChannelisedVoltageStream(AntennaChannelisedVoltageStreamBase):
...
         if n_chans < 1 or (n_chans & (n_chans - 1)) != 0:
             raise ValueError('n_chans is not a power of 2')
         if len(src_streams) % 2 != 0:
             raise ValueError('src_streams does not have an even number of elements')
+        self.input_labels = (
+            [stream.name for stream in src_streams]
+            if input_labels is None else list(input_labels)
+        )
+        if len(self.input_labels) != len(src_streams):
+            raise ValueError('input_labels has the wrong number of elements')
```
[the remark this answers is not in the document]
bmerry: Fair enough, will do.

katsdpcontroller/product_config.py
bmerry: Yes, integration time. IIRC that spelling is part of the CBF-CAM ICD.

katsdpcontroller/test/test_product_config.py
bmerry: Firstly katsdpcontroller uses nosetests rather than pytest (because it's old - I'd like to port it one day). But I don't think this is the right place for xfail: xfail is for tests that you would really like to pass but the don't yet (maybe due to a bug you can't immediately fix). If this was pytest I'd use `pytest.raises`.

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
bmerry: Done.

james-smith-za (approved):
LGTM.

katsdpcontroller/test/test_product_config.py
james-smith-za: Understood.

katsdpcontroller/product_config.py
james-smith-za: I guess it would help if I read the document! 😊

katsdpcontroller/product_config.py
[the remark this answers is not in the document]
james-smith-za: All fair points. Probably easiest to leave it as you've done it.

PR #535, opened 2021-07-15
Adjust resource requirements for fgpu and dsim

bmerry:
Memory usage is estimated from running examples. Compute is in some
places scaled according to the ADC sample rate, so that low-bandwidth
simulations will more easily fit on a laptop in future.

See NGC-218.


james-smith-za (approved):
Is there something special about the number 3072? Do you make them both the same on purpose?

bmerry:
> Is there something special about the number 3072? 

It's a nice round number (0xc00).

PR #537, opened 2021-07-23
Pass --ibv to dsim when not in develop mode

bmerry:
This is to complement https://github.com/ska-sa/katgpucbf/pull/35 and
should be merged at the same time. When develop mode is not set, pass
--ibv to dsim (new) and do all the normal ibverbs things. When develop
mode is set, don't try to use ibverbs.

See NGC-200.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -266,6 +266,7 @@ def _make_dsim(
     An antenna has a separate stream per polarisation, so `streams` will
     normally have two elements.
     """
+    ibv = not configuration.options.develop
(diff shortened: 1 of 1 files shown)


james-smith-za (approved):
My question would have been whether you only anticipate using "develop mode" on your local laptop, and never on an actual server where you  might want to use ibv functionality.

But after reading through the rest of the code, this seems to be an established thing for this module. If I've understood correctly.

Looks good to me.

bmerry:
> My question would have been whether you only anticipate using "develop mode" on your local laptop, and never on an actual server where you might want to use ibv functionality.

I guess it is a bit of a one-size-fits-all switch - but it's worked well enough up to now.

PR #538, opened 2021-07-26
Copy old aiokatcp.test.test_utils.timelimit into repo

bmerry:
In aiokatcp 1.0, the aiokatcp.test.test_utils module was removed
(because aiokatcp switched to pytest). We relied on the `timelimit`
decorator, so copy-and-paste it into katsdpcontroller.

Closes SPR1-1317.

### DIFF
--- a/katsdpcontroller/test/utils.py
+++ b/katsdpcontroller/test/utils.py
@@ -1,6 +1,9 @@
 """Utilities for unit tests"""
 
 import asyncio
+import functools
+import inspect
+import sys
 import unittest
 from unittest import mock
 from typing import (List, Tuple, Iterable, Coroutine, Awaitable, Optional,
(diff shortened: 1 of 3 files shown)


katsdpcontroller/test/utils.py
```diff
@@ -369,3 +373,39 @@ async def replacement(*args, **kwargs):
...
+            async def wrapper(self, *args, **kwargs):
+                try:
+                    async with async_timeout.timeout(limit, loop=self.loop) as cm:
+                        await arg(self, *args, **kwargs)
+                except asyncio.TimeoutError:
+                    if not cm.expired:
+                        raise
+                    for task in asyncio.Task.all_tasks(loop=self.loop):
+                        if task.get_stack(limit=1):
+                            print()
```
richarms, line 406: Not sure what this print() statement is doing?

richarms (approved):

katsdpcontroller/test/utils.py
bmerry: It puts a blank line between the stacks of the tasks so that one can tell where the boundaries are - otherwise it all just looks like one big stack trace.

PR #539, opened 2021-07-29
Support image_overrides in product-config schema

bmerry:
This provides an equivalent to the `--image-override` command-line
option, but on a per-product basis.

Closes NGC-273.

### DIFF
--- a/katsdpcontroller/product_config.py
+++ b/katsdpcontroller/product_config.py
@@ -208,10 +208,12 @@ class Options:
     def __init__(self, *, develop: bool = False,
                  wrapper: Optional[str] = None,
                  image_tag: Optional[str] = None,
+                 image_overrides: Mapping[str, str] = {},
                  service_overrides: Mapping[str, ServiceOverride] = {}) -> None:
         self.develop = develop
         self.wrapper = wrapper
         self.image_tag = image_tag
+        self.image_overrides = dict(image_overrides)
         self.service_overrides = dict(service_overrides)
 
     @classmethod
@@ -224,6 +226,7 @@ class Options:
             develop=config.get('develop', False),
             wrapper=config.get('wrapper'),
             image_tag=config.get('image_tag'),
+            image_overrides=config.get('image_overrides', {}),
             service_overrides=service_overrides
         )
 
--- a/katsdpcontroller/product_controller.py
+++ b/katsdpcontroller/product_controller.py
@@ -1491,8 +1491,11 @@ class DeviceServer(aiokatcp.DeviceServer):
             resolver_factory_args = dict(tag=image_tag)
         else:
             resolver_factory_args = {}
+        image_resolver = self.image_resolver_factory(**resolver_factory_args)
+        for image_name, image_path in configuration.options.image_overrides.items():
+            image_resolver.override(image_name, image_path)
         resolver = Resolver(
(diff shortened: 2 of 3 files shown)


bmerry:
An example of use:
```
{
    "version": "3.1",
    "config": {
        "develop": true,
        "image_overrides": {
            "katgpucbf": "localhost:5000/katgpucbf@sha256:2903d5ce2f98ab13a0a1b987e4a64524d38e845e7f98c72bc61fa37715a4997c"
        }
    },
    ....
}
```

james-smith-za (commented):
Am I understanding correctly that you would use this option to say, test a specific new F-engine while keeping everything else the same?

I'm used to identifying container images with something like `katgpucbf:v0.2.3`, not a sha256 hash. How would one go about getting that information in order to specify the specific image to use?

bmerry:
> Am I understanding correctly that you would use this option to say, test a specific new F-engine while keeping everything else the same?

It doesn't support overriding just the F-engines; it'll override everything that uses the katgpucbf Docker image.

> I'm used to identifying container images with something like katgpucbf:v0.2.3, not a sha256 hash

You can do that too, but for reasons I don't recall, if there is an older version of a tag on the machine where it runs, it won't update it. A hash means you can be sure of what you're getting. There are a few ways to get the hash e.g. I think `docker pull` will report it.

james-smith-za:
I'm just trying to understand the reason for this. The `--image-tag` facility exists, but you're providing the facility to completely override the image (and tag) to use.

So on your machine you may have a release of the `katgpucbf` image which came from the docker registry, but you've built a new one locally that you'd like to test, so you override with that?

Is that closer to the purpose you had in mind?

bmerry:
> I'm just trying to understand the reason for this. The --image-tag facility exists, but you're providing the facility to completely override the image (and tag) to use.

The `image_tag` (either on the command-line or in the JSON) sets the tag for ALL images. So if I say `image_tag: bmerry`, then I have to ensure there is a bmerry tag of katsdpcontroller, katgpucbf, katsdpingest, katsdpmetawriter, katsdpcam2telstate etc etc. There is a utility on the SDP Jenkins server to construct such tags (by copying or mostly copying an existing tag, for all images), but it's an extra manual step. With `image_override` I can change the tag for a SINGLE image e.g. use katgpucbf:my-test-branch while still using the `latest` tag for all the other images.

It's also possible to override the registry, but that's more incidental. It could come in handy for lab work if we have a registry in the lab, so that we don't have to push images to site and then have nodes pull things back from site inside the code-build-test loop.

james-smith-za:
> The `image_tag` (either on the command-line or in the JSON) sets the tag for ALL images. So if I say `image_tag: bmerry`, then I have to ensure there is a bmerry tag of katsdpcontroller, katgpucbf, katsdpingest, katsdpmetawriter, katsdpcam2telstate etc etc. There is a utility on the SDP Jenkins server to construct such tags (by copying or mostly copying an existing tag, for all images), but it's an extra manual step. With `image_override` I can change the tag for a SINGLE image e.g. use katgpucbf:my-test-branch while still using the `latest` tag for all the other images.

Thanks. That puts things into better context now - I was unaware of this relationship between all the katsdp packages. But given that restriction, the ability to override *this* image tag makes perfect sense.

james-smith-za (approved):

PR #540, opened 2021-08-03
xbgpu now uses katsdpservices for logging

bmerry:
See NGC-288.


bmerry:
Note: untested. Once this is merged in katgpucbf let me know so that I can verify that it's working (it looks like it might not be on fgpu, but fgpu is very quiet by default).

james-smith-za (approved):
Seems reasonable enough. I am not in a position to test this myself, but I won't let that hold things up.

If things get stuck later I will open a bug on JIRA.

PR #547, opened 2021-09-08
Improve user experience with image overrides

bmerry:
This has three parts:
1. Put image overrides through the lookup machinery
2. Allow config image_overrides to override the product controller
3. Some small code cleanups.

See the individual commits for more details.

Closes SPR1-1466.


richarms (approved):
Took a while to verify this. All appears to work as intended/described (i.e. mesos agent does not use older version of specified image if just NAME:TAG is supplied instead of image hash, and full path not required)

PR #552, opened 2021-09-29
Support token authentication with a registry

bmerry:
This will allow katsdpcontroller to work with harbor as the registry (or
any registry implementing
https://docs.docker.com/registry/spec/auth/token/). When an HTTP 401
response is receive with a suitable WWW-Authenticate, send another
request to that token server to get a token and then retry the original
request with that token.

This also contains a bug fix: if an explicit registry is provided for a
request, look up the appropriate username and password for that
registry, rather than using the credentials for the default registry.

The unit test has also been significantly beefed up, to test a number of
failure cases, and to actually check that credentials are being looked
up and passed correctly (which also required fixing the test to store
them in the right format).

Closes NGC-389.

### DIFF
--- a/katsdpcontroller/scheduler.py
+++ b/katsdpcontroller/scheduler.py
@@ -182,7 +182,7 @@ import jsonschema
 from decorator import decorator
 from addict import Dict
 import pymesos
-
+import www_authenticate
 import aiohttp.web
 
 from katsdptelstate.endpoint import Endpoint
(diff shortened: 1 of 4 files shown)


katsdpcontroller/scheduler.py
```diff
@@ -944,20 +992,32 @@ async def __call__(self, repo: str, tag: str) -> str:
...
-                    digest = response.headers['Docker-Content-Digest']
-            except (aiohttp.client.ClientError, asyncio.TimeoutError) as error:
-                raise ImageError('Failed to get digest from {}: {}'.format(url, error)) \
-                    from error
-            except KeyError:
-                raise ImageError('Docker-Content-Digest header not found for {}'.format(url))
+                auth_header = auth.encode() if auth else None
+                digest = await self._get_digest(session, url, ssl_context, auth_header)
+            except ImageError as error:
+                cause = error.__cause__
```
martinslabber, line 1000: This is a little gem. I was not aware of context coming through with exception like this.

martinslabber (approved):
I did not fine come the most interesting part, the test.

PR #554, opened 2021-10-18
MKAT-108 Pass excise flag from product to ingest

ludwigschwardt:
We want to be able to disable excision, but the instruction was lost somewhere between the product controller and the ingest node.


ludwigschwardt:
@bmerry, please check that this is sufficient :-)

PR #558, opened 2021-11-03
Decrease number of substreams for gpucbf correlator

bmerry:
Some updates were needed to the memory usage calculation too - see the second commit. Testing with 4 antennas and either 1024 or 32768 channels shows actual memory usage is 45-60% of that allocated. We can always trim it a bit closer later if necessary, but I'd rather see how things scale with the number of antennas.

See NGC-354.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -527,13 +527,22 @@ def _make_xbgpu(
         g.graph["static_sensors"].add(ss)
 
     bw_scale = stream.adc_sample_rate / _MAX_ADC_SAMPLE_RATE
-    # * 4 is for 2 polarisations, each with real+complex
-    batch_size = len(acv.antennas) * stream.n_chans_per_endpoint * 4 * acv.bits_per_sample // 8
+    # * 2 is for real+complex
+    batch_size = (
+        len(acv.src_streams) * acv.n_spectra_per_heap
+        * stream.n_chans_per_endpoint * 4 * acv.bits_per_sample // 8
+    )
+    rx_reorder_tol = 536870912  # Default of --rx-reorder-tol option
+    # Memory allocated for buffering and reordering incoming data
+    # (no *2 for real+complex here: it cancels out).
+    recv_buffer = round(
+        rx_reorder_tol * len(acv.src_streams) * acv.bits_per_sample / 8 / stream.n_substreams
+    )
     for i in range(0, stream.n_substreams):
         xbgpu = SDPLogicalTask(f'xb.{stream.name}.{i}')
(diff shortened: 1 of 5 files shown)


katsdpcontroller/product_config.py
```diff
@@ -530,11 +530,14 @@ def __init__(
...
-        # F-engine, rounded up to a power of two.
+        # Determine how fine to divide the stream, i.e., the number of xgpu
+        # processes to run. The constant is the data rate (in bytes per
+        # second) that one engine can handle, and is sufficient for S-band
+        # with 80 antennas to be handled by 64 engines.
+        # The minimum is 4 since SDP expects to run 4 ingest processes.
+        first.adc_sample_rate
         n_substreams = 4
-        while n_substreams < 4 * (len(src_streams) // 2):
+        while n_substreams * 4.375e9 < first.adc_sample_rate * len(src_streams):
```
james-smith-za, line 540: I'd just suggest making this more explicit by giving this constant a name:
```
max_datarate_per_engine: Final = 4.375e9
```
I know you've referred to it in the comment, but I can imagine the number (or even the comparison) being modified without the comment being updated.

katsdpcontroller/generator.py
```diff
@@ -527,14 +527,23 @@ def get_baseline_index(a1, a2):
         g.graph["static_sensors"].add(ss)
 
     bw_scale = stream.adc_sample_rate / _MAX_ADC_SAMPLE_RATE
-    # * 4 is for 2 polarisations, each with real+complex
-    batch_size = len(acv.antennas) * stream.n_chans_per_endpoint * 4 * acv.bits_per_sample // 8
+    # * 2 is for real+complex
+    batch_size = (
+        len(acv.src_streams) * acv.n_spectra_per_heap
+        * stream.n_chans_per_endpoint * 4 * acv.bits_per_sample // 8
+    )
```
james-smith-za, line 534: Comment doesn't seem to match up with the actual assignment.

james-smith-za (changes requested):
Mostly bike-shedding from me but I think worth your consideration.

katsdpcontroller/generator.py
bmerry: Thanks - I assume you're referring to the `* 4` instead of the `* 2`. I've fixed it, and now memory usage is 60-70% of allocated, which feels comfortable.

katsdpcontroller/product_config.py
bmerry: Good point - I've added a constant to defaults.py where some similar constants live.

james-smith-za (approved):

PR #560, opened 2021-11-08
Use katsdptelstate.aio.redis.RedisBackend.from_url

bmerry:
This removes the direct dependency on aioredis, and will ease the
transition to aioredis2.

Depends on https://github.com/ska-sa/katsdptelstate/pull/125

### DIFF
--- a/katsdpcontroller/product_controller.py
+++ b/katsdpcontroller/product_controller.py
@@ -19,7 +19,6 @@ import aiokatcp
 from aiokatcp import FailReply, Sensor, Address
 from prometheus_client import Gauge, Counter, Histogram, CollectorRegistry, REGISTRY
 import yarl
-import aioredis
 import katsdptelstate.aio.redis
 import katsdpmodels.fetch.aiohttp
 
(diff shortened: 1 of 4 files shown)


katsdpcontroller/product_controller.py
```diff
@@ -1172,8 +1171,9 @@ async def _launch_telstate(self) -> katsdptelstate.aio.TelescopeState:
         # connect to telstate store
         self.telstate_endpoint = '{}:{}'.format(self.telstate_node.host,
                                                 self.telstate_node.ports['telstate'])
-        redis_client = await aioredis.create_redis_pool(f'redis://{self.telstate_endpoint}')
-        telstate_backend = katsdptelstate.aio.redis.RedisBackend(redis_client)
+        telstate_backend = await katsdptelstate.aio.redis.RedisBackend.from_url(
```
ludwigschwardt, line 1174: Why construct the backend at all? Why not just pass the URL to `TelescopeState`?

ludwigschwardt (commented):
I've probably forgotten how this all works, hence the silly questions :-)

katsdpcontroller/product_controller.py
bmerry: The aio version of TelescopeState can only be constructed from a Backend - something I did to avoid the complex overloading in the synchronous version. That also solves the problem that a constructor cannot be asynchronous.

ludwigschwardt: Right, thanks!

ludwigschwardt (approved):

PR #561, opened 2021-11-09
Tweaks to linkage with consul

bmerry:
The consul `Name` is changed to be the task type: this means that consul
service discovery can be used to find all F-engine tasks for example.
The specific instance e.g. `f.antenna_channelised_voltage.1` is put into
a `task_name` metadata field.

The sandbox Prometheus config is also adjusted: the task type is now
mapped to `job` (a standard Prometheus label).

Relates to NGC-325.


richarms (approved):

bmerry:
@james-smith-za I'm going to merge this based on Richard's approval, because there are a few PRs built on top of this. You're still welcome to comment if you wish.

PR #564, opened 2021-11-12
Implement ?delays request

bmerry:
There is no proxying via the master controller - that's TBD depending on
what the ICD ends up looking like.

Closes NGC-411.

### DIFF
--- a/katsdpcontroller/product_controller.py
+++ b/katsdpcontroller/product_controller.py
@@ -363,6 +363,10 @@ class SDPSubarrayProductBase:
     conditions, changes to :attr:`state` should generally only be made from
     inside the asynchronous tasks.
 
+    Some methods that are asynchronous but don't change state (such as
+    setting delays) may run concurrently. They are best-effort, and may fail if
+    the subarray product is deconfigured concurrently.
+
     There are some invariants that must hold at yield points:
     - There is at most one capture block in state CAPTURING.
     - :attr:`current_capture_block` is the capture block in state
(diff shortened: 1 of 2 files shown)


katsdpcontroller/product_controller.py
```diff
@@ -1329,6 +1380,26 @@ def must_wait(node):
...
+            stream: product_config.GpucbfAntennaChannelisedVoltageStream,
+            timestamp: aiokatcp.Timestamp,
+            coefficient_sets: Sequence[str]) -> None:
+        n_inputs = len(stream.src_streams)
+        conns = []
+        for i in range(n_inputs // 2):
+            node = self._nodes[f"f.{stream.name}.{i}"]
+            if node.katcp_connection is None:
+                raise FailReply(f"No katcp connection to {node.name}")
+            conns.append(node.katcp_connection)
```
james-smith-za, line 1394: So the one reservation that I'd have with this, is that if a single f-host wasn't connected, then the entire `?delays` request fails. Is that behaviour that we want?

james-smith-za (commented):
Code looks ok to me, just one query regarding the actual behaviour that we want to implement.

I don't see any specification written one way or the other in the ICD, and this implementation may be ok for now, but it's probably something that we are going to want to come back to later.

katsdpcontroller/product_controller.py
bmerry: I think it's reasonable until we have health sensors. If we're unable to communicate a delay model to an engine, it should be marked as unhealthy (and maybe eventually killed and restarted).

bmerry:
@james-smith-za gave verbal approval, so merging.

PR #569, opened 2021-12-09
Customise sensor name mangling for fgpu

bmerry:
- Add generic mechanism to SensorProxyClient to allow sensor renaming to
  be overridden for specific sensors.
- Add hook in SDPPhysicalTask to allow the logical task to specify the
  renaming.
- Add renaming rules for `gains` and `delay` for fgpu.

Closes NGC-451.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -461,6 +461,11 @@ def _make_fgpu(
         # Link it to the group, so that downstream tasks need only depend on the group.
         g.add_edge(fgpu_group, fgpu, depends_ready=True, depends_init=True)
 
+        # Rename sensors that are relevant to the stream rather than the process
+        for j, src in enumerate(srcs):
+            fgpu.sensor_renames[f"input{j}-eq"] = f"{stream.name}-{src.name}-eq"
+            fgpu.sensor_renames[f"input{j}-delay"] = f"{stream.name}-{src.name}-delay"
+
     return fgpu_group
 
 
--- a/katsdpcontroller/sensor_proxy.py
+++ b/katsdpcontroller/sensor_proxy.py
@@ -43,9 +43,11 @@ class SensorWatcher(aiokatcp.SensorWatcher):
     def __init__(self, client: aiokatcp.Client, server: aiokatcp.DeviceServer,
                  prefix: str,
                  rewrite_gui_urls: Callable[[aiokatcp.Sensor], bytes] = None,
-                 enum_types: Sequence[Type[enum.Enum]] = ()) -> None:
+                 enum_types: Sequence[Type[enum.Enum]] = (),
+                 renames: Optional[Mapping[str, str]] = None) -> None:
         super().__init__(client, enum_types)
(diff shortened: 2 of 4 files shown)


bmerry:
FYI @amishatishpatel this should be ready for the katsdpcontroller side of NGC-436.

bmerry:
Agreed that @amishatishpatel will review since James is away.

katsdpcontroller/generator.py
```diff
@@ -461,6 +461,11 @@ def _make_fgpu(
         # Link it to the group, so that downstream tasks need only depend on the group.
         g.add_edge(fgpu_group, fgpu, depends_ready=True, depends_init=True)
 
+        # Rename sensors that are relevant to the stream rather than the process
+        for j, src in enumerate(srcs):
+            fgpu.sensor_renames[f"input{j}-eq"] = f"{stream.name}-{src.name}-eq"
+            fgpu.sensor_renames[f"input{j}-delay"] = f"{stream.name}-{src.name}-delay"
```
amishatishpatel, line 467: I see, I see. It really is name *mangling*!

katsdpcontroller/sensor_proxy.py
```diff
@@ -59,7 +61,7 @@ def __init__(self, client: aiokatcp.Client, server: aiokatcp.DeviceServer,
         self._interface_changed = False
 
     def rewrite_name(self, name: str) -> str:
-        return self.prefix + name
+        return self.renames.get(name, self.prefix + name)
```
amishatishpatel, line 64: I understand that it will only ever receive a `sensor.name` for one that is correctly instantiated, but if I understand this correctly:
* This `get` call will return `self.prefix + name` whether the name is present or not.

Is that/should it be a concern?

image

katsdpcontroller/test/test_sensor_proxy.py
```diff
@@ -94,6 +95,8 @@ def _check_sensors(self) -> None:
         """Compare the upstream sensors against the mirror"""
         for sensor in self.server.sensors.values():
             qualname = 'prefix-' + sensor.name
+            if sensor.name == 'bytes-sensor':
+                qualname = 'custom-bytes-sensor'
```
amishatishpatel, line 99: Any particular reason this one needed to be updated?

amishatishpatel (commented):
Looks good to me, just a few small questions before I can approve.

katsdpcontroller/sensor_proxy.py
bmerry: It'll return `self.renames[name]` if that exists, otherwise `self.prefix + name`. So anything without an explicit override will default to the old behaviour.

katsdpcontroller/test/test_sensor_proxy.py
bmerry: Because of the `renames` passed at line 88. I picked one arbitrarily to give the renaming treatment.

amishatishpatel (approved):

PR #575, opened 2022-02-23
Reserve a separate core for the fgpu Python thread

bmerry:
This becomes necessary with smaller chunk sizes, which push more work
onto the Python thread.

Relates to NGC-505.


james-smith-za (approved):

james-smith-za:
If we are doing 4x fgpu per host, and using 4 cores per instance, then we're maxing out our available CPU cores...

bmerry:
> If we are doing 4x fgpu per host, and using 4 cores per instance, then we're maxing out our available CPU cores...

Yep. Or as I think of it, preventing any pesky interfering processes from being scheduled at the same time.

PR #576, opened 2022-03-18
Enable aiomonitor/aioconsole with fgpu/xbgpu

bmerry:
This will require changes in the katgpucbf repo to make it work, but
should aid in debugging.

### DIFF
--- a/katsdpcontroller/generator.py
+++ b/katsdpcontroller/generator.py
@@ -409,7 +409,8 @@ def _make_fgpu(
         fgpu.cores = ['src0', 'src1', 'dst', 'python']
         if not configuration.options.develop:
             fgpu.numa_nodes = 1.0  # It's easily starved of bandwidth
-        fgpu.ports = ['port', 'prometheus']
+        fgpu.ports = ['port', 'prometheus', 'aiomonitor', 'aioconsole']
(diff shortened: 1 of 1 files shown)


katsdpcontroller/generator.py
```diff
@@ -409,7 +409,8 @@ def _make_fgpu(
         fgpu.cores = ['src0', 'src1', 'dst', 'python']
         if not configuration.options.develop:
             fgpu.numa_nodes = 1.0  # It's easily starved of bandwidth
-        fgpu.ports = ['port', 'prometheus']
+        fgpu.ports = ['port', 'prometheus', 'aiomonitor', 'aioconsole']
```
james-smith-za, line 412: Given all these new ports, it may be worth renaming `port` to something like `katcp`, unless there's already a convention on katsdpcontroller which should be followed.

james-smith-za (approved):

katsdpcontroller/generator.py
bmerry: It's already a convention in katsdpcontroller. With SDP tasks the port mapping are written into the telstate config for the task and all the SDP containers look for it under that name, so changing it would be a pretty major exercise that would need lockstep changes across multiple repos.

It is a naming choice I regret though.

PR #581, opened 2022-03-26
Pin xbgpu to cores[dst]

bmerry:
This will prevent the Python thread (and any auxiliary threads e.g.
those started by CUDA) from taking CPU time away from the
performance-critical receiver thread. Hopefully this will reduce the
incidences of lost packets.


bmerry:
Unfortunately it doesn't seem to have prevent packets getting dropped, but it's still worth fixing.

james-smith-za (approved):

PR #582, opened 2022-03-31
Allow a network interface to belong to multiple networks

bmerry:
In the Mesos attribute it is permitted to take a list instead of a
single entry.

### DIFF
--- a/katsdpcontroller/scheduler.py
+++ b/katsdpcontroller/scheduler.py
@@ -797,9 +797,7 @@ class GPURequest(ResourceRequestsContainer):
 
 
 class InterfaceRequest(ResourceRequestsContainer):
-    """Request for resources on a network interface. At the moment only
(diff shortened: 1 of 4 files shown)


bmerry:
This is to support running SDP and CBF in the lab using the same physical network.

katsdpcontroller/scheduler.py
```diff
@@ -797,9 +797,7 @@ def __repr__(self):
 
 
 class InterfaceRequest(ResourceRequestsContainer):
-    """Request for resources on a network interface. At the moment only
-    a logical network name can be specified, but this may be augmented in
-    future to allocate bandwidth.
+    """Request for resources on a network interface.
```
bmerry, line 800: This change isn't actually related to the PR, but I noticed this comment was very out of date.

PR #583, opened 2022-03-31
Don't set core affinities for fgpu/xbgpu with --develop

bmerry:
The affinities prevent scaling down a correlator to fit on a laptop when
using a low sample rate.


james-smith-za (approved):
LGTM. Though AFAIK you're the only one who uses it on a laptop, I don't know who else it'll affect.

PR #586, opened 2022-04-05
Redesign interface mode to operate at lower level

bmerry:
Up to now, `--interface-mode` has worked by having two subclasses of SDPSubarrayProductBase, one of which does nothing. That meant a significant amount of business logic (including creation of a number of sensors) did not run in interface mode and the fidelity of the simulation was limited.

To improve that, the split between production and simulation is pushed down to the scheduler. A SchedulerBase class is extracted which does not require a Mesos server. It is unable to launch instances of PhysicalTask, but a FakePhysicalTask substitute class is provided. The generator substitutes this in where appropriate, and provides subclasses for more detailed emulation of particular containers. This gives a much larger set of simulated sensors, and one that depends on the configuration of the subarray product rather than being hard-coded. Postprocessing is not well-emulated; that would be difficult because the tasks to launch depend on querying data from the observation.

I've also sprinkled on some more type annotations to help me identify bugs. It's certainly not complete.

I've tried to keep the changes relatively minimal without breaking functionality. There is further cleanup that could be done in a future PR. In particular, SDPSubarrayProductBase and SDPSubarrayProduct can be merged. A lot of controller unit tests can probably be more easily performed in interface mode, although care would be needed to avoid reducing coverage, particularly for telstate config callbacks in generator.py.

Relates to NGC-498, although it does not directly address any CBF functionality, and in fact it will go backwards as commands like `?gain` are not accurately emulated.


bmerry:
@james-smith-za FYI, but since it's mostly affecting SDP functionality I've assigned it to @richarms.

james-smith-za:
> @james-smith-za FYI, but since it's mostly affecting SDP functionality I've assigned it to @richarms.

Had a quick look, can't see any glaring problems - but I am not really familiar enough with the code base to make informed comment. Thanks for the efforts, Bruce.

bmerry:
@richarms when do you think you'll have time to review this?

richarms:
Apologies for the delay, this one managed to pass me by until being reminded. What's been done makes sense logically, and I can't see any obvious problems, nor anticipate downstream issues for SDP. I'm happy to accept all changes.

richarms (approved):

PR #587, opened 2022-04-20
Implement fake versions of a number of katgpucbf requests

bmerry:
- Implement `gain` and `delay` requests in interface mode.
- Implement `gain-all` support (all modes), although fgpu doesn't implement it yet so it won't actually work.
- Implement `capture-start` and `capture-stop`.
- Implement PFB group delay sensor

Relates to NGC-498.


katsdpcontroller/generator.py
```diff
@@ -461,6 +407,9 @@ def _make_fgpu(
         Sensor(int, f"{stream.name}-n-samples-between-spectra",
                "Number of samples between spectra",
                default=stream.n_samples_between_spectra, initial_status=Sensor.Status.NOMINAL),
+        Sensor(int, f"{stream.name}-pfb-group-delay",
+               "PFB group delay, specified in number of samples",
+               default=-stream.n_chans * stream.pfb_taps, initial_status=Sensor.Status.NOMINAL),
```
james-smith-za, line 412: I'm still not convinced that this is a correct use of the term "group delay" which as far as I understand it, means the time delay of a signal through a device as a function of frequency. But if that's the term used higher up on the system level then there's little point I guess in being stubborn about it.

I would just want to make sure that `n_chans * pfb_taps` is actually the way that it's calculated on a system level, I've never actually gotten round to getting clarity on that.

katsdpcontroller/generator.py
```diff
@@ -493,8 +442,9 @@ def _make_fgpu(
 
     for i in range(0, n_engines):
         srcs = stream.sources(i)
-        fgpu = SDPLogicalTask(f'f.{stream.name}.{i}')
+        fgpu = SDPLogicalTask(f'f.{stream.name}.{i}', streams=[stream])
         fgpu.image = 'katgpucbf'
+        fgpu.fake_katcp_server_cls = FakeFgpuDeviceServer
```
james-smith-za, line 447: Is there not an equivalent of this for xbgpu?

katsdpcontroller/product_controller.py
```diff
@@ -907,6 +899,43 @@ async def gain_impl(
...
 
+    async def gain_all(self, stream_name: str, values: Sequence[str]) -> None:
+        """Set F-engine gains for all inputs."""
+        if self.state not in {ProductState.CAPTURING, ProductState.IDLE}:
+            raise FailReply(f"Cannot set gains in state {self.state}")
+        stream = self._find_stream(stream_name)
+        if not isinstance(stream, product_config.GpucbfAntennaChannelisedVoltageStream):
+            raise FailReply(f"Stream {stream_name!r} is of the wrong type")
+        if len(values) not in {1, stream.n_chans}:
+            raise FailReply(f"Expected 0, 1, or {stream.n_chans} values, received {len(values)}")
```
james-smith-za, line 910: Maybe I'm missing something, but your text says you expect possibly 0, but you don't check for that in the if-statement.

katsdpcontroller/product_controller.py
```diff
@@ -907,6 +899,43 @@ async def gain_impl(
...
+        if values != ("default",):
+            try:
+                for v in values:
+                    complex(v)  # Evaluated just for the exception
+            except ValueError as exc:
+                raise FailReply(str(exc))
+        # TODO: remove the type: ignores once SDPSubarrayProduct and SDPSubarrayProductBase merged
+        await self._multi_request(  # type: ignore
+            self.find_nodes(task_type="f", streams=[stream]),  # type: ignore
+            itertools.repeat(("gain-all",) + values)
```
james-smith-za, line 921: Do the f-engines need to know about the gain-all request to the PC? My understanding was that the PC would just translate the gain-all request to individual (identical) gain requests for each F-engine. Or maybe I'm misunderstanding.

```suggestion
            itertools.repeat(("gain",) + values)
```

Edit: I think I've just understood, in that `?gain` needs the name of the input as well, which in the case of the F-engine will (I guess) correspond to the polarisation being adjusted.

So my suggestion won't work - either we'd need to know the names of the inputs that each F-engine has, or to implement a name-agnostic `?gain-all` at the F-engine which I think was your original intention.

katsdpcontroller/product_controller.py
```diff
@@ -907,6 +899,43 @@ async def gain_impl(
...
+        )
+
+    async def capture_start_stop(self, stream_name: str, start: bool) -> None:
+        """Either start or stop transmission on a stream."""
```
james-smith-za, line 925: On the one hand I'm somewhat uncomfortable with a function that starts or stops depending on the bool that gets passed in.

On the other hand, I think I can see why you did this, because there would be a lot of repetition if you had separate functions for each, since they're both basically the same.

As a compromise, I'd suggest putting a `*` in the argument list before the `start` parameter to force the caller to use a named argument. That way when I see
`capture_start_stop("my_stream", start=False)`, then I know it's not starting so it must be stopping.

Alternatively a pattern that I've seen used is to have something like 
```python
def capture_start_stop(action: str):
    if action not in {"start", "stop"}:
        raise ValueError("Expected `start` or `stop`")
    else:
        ...
```

I may just be bike-shedding at this point though. It's difficult to tell.

james-smith-za (commented):
The Jenkins failure is concerning but it seems to have been a katpoint build issue, not anything to do with this PR - perhaps it just needs a kick?

bmerry:
> The Jenkins failure is concerning but it seems to have been a katpoint build issue, not anything to do with this PR - perhaps it just needs a kick?

One of the build machines filled its hard drive and everything was failing. I'll check it against after I make updates.

katsdpcontroller/product_controller.py
bmerry: I like the named argument idea. An alternative would be to use an Enum (better than a string since mypy can do the type check statically), but that feels overkill for a single use.

katsdpcontroller/product_controller.py
bmerry: It could issue two `gain` commands (one for each pol) without too much trouble, but that would double the network bandwidth, wouldn't work so well with `_multi_request`, and it wouldn't handle the magic `default` value. I figure it will be better to just implement `gain-all` within fgpu.

katsdpcontroller/product_controller.py
bmerry: Error in the comment - thanks (copy-pasted from `?gain`, which does support passing 0 values to query).

katsdpcontroller/generator.py
bmerry: There probably will be eventually, if it needs to expose sensors. The default is `FakeDeviceServer` which is good enough to handle requests for which there is no reply, such as `capture-start` / `capture-stop`.

katsdpcontroller/generator.py
bmerry: I've done some tracing through the various tickets, and the goal seems to be that if you have an impulse at timestamp 0, what is the output timestamp of the centroid of the response (modulo the negative sign). Since the PFB filter is symmetric and of length n_chans * n_taps * 2, the centroid is at n_chans * n_taps (actually I think it should be n_chans * n_taps - 0.5, but it's specified to be integer). I've also confirmed that this is what MeerKAT is exposing in a few different instruments.

ludwigschwardt: To add to the confusion... :grin:

"Group delay" is the local phase slope at a given frequency. It can therefore vary across the band if you have a wobbly phase response.
"Phase delay" is the "global" phase slope obtained by connecting the phase response at a given frequency with the origin.

Since the PFB is a FIR filter, it has linear phase. This makes the above distinctions moot, since the phase response is a straight line through the origin. Therefore, group delay = phase delay = "delay".

PR #589, opened 2022-04-22
Merge SDPSubarrayProductBase into SDPSubarrayProduct

bmerry:
After #586 there is no longer a need for this abstraction layer. There should be no changes in semantics, just merging the classes together and inlining previously abstract functions.

### DIFF
--- a/katsdpcontroller/product_controller.py
+++ b/katsdpcontroller/product_controller.py
@@ -527,21 +527,8 @@ class SDPSubarrayProduct:
         if self.telstate is None:
             raise FailReply(f'Subarray product {self.subarray_product_id} has no SDP components.')
 
-    async def capture_init_impl(self, capture_block: CaptureBlock) -> None:
-        """Extension point to start a capture block.
-
-        If it raises an exception, the capture block is assumed to not have
-        been started, and the subarray product goes into state ERROR.
-        """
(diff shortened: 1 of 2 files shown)


katsdpcontroller/product_controller.py
```diff
@@ -442,10 +458,28 @@ def __init__(self, sched: scheduler.SchedulerBase,
...
         logger.info("Created: %r", self)
         logger.info('Logical graph nodes:\n'
                     + '\n'.join(repr(node) for node in self.logical_graph))
 
+    def __del__(self) -> None:
+        if hasattr(self, 'batch_queue'):
+            self.sched.remove_queue(self.batch_queue)
+
+    def __repr__(self) -> str:
+        return "Subarray product {} (State: {})".format(self.subarray_product_id, self.state.name)
```
richarms, line 481: Seems as good a time as any to replace last few remaining `str.format` with f-strings :)

richarms (approved):

katsdpcontroller/product_controller.py
bmerry: True. I can run pyupgrade on it, but I'll make a separate PR for that since it'll generate a lot of simple changes that are a pain to review when mixed in with functional changes.

PR #591, opened 2022-05-09
Updates to reflect that it's not just for SDP

bmerry:
The controller is no longer SDP-specific, being used for the new CBF as well. Update it to remove the "SDP" term from a number of classes, docstrings, web page titles etc. This should have no functional impact, and is just designed to reduce programmer and user confusion.

I'm not totally happy with changing `tasks.SDPLogicalTask` to `tasks.LogicalTask` because it now has the same base name as `scheduler.LogicalTask` (the same applies to PhysicalTask etc). If anyone has a better name to reflect that this is the version with extra sauce that is specific to the use case within the product controller rather than suitable for a general-purpose scheduler I'm open to suggestions.

Closes NGC-546.

### DIFF
--- a/katsdpcontroller/dashboard.py
+++ b/katsdpcontroller/dashboard.py
@@ -19,7 +19,7 @@ import dash_table
 from dash_dangerously_set_inner_html import DangerouslySetInnerHTML
 
 from . import scheduler
-from .tasks import PhysicalTask
+from .tasks import ProductPhysicalTask
 
 
 def timestamp_utc(timestamp):
@@ -45,7 +45,7 @@ def _get_tasks(product):
     order_graph = scheduler.subgraph(product.physical_graph, scheduler.DEPENDS_READY)
     tasks = networkx.lexicographical_topological_sort(
             order_graph.reverse(), key=lambda node: node.name)
-    tasks = [task for task in tasks if isinstance(task, PhysicalTask)]
+    tasks = [task for task in tasks if isinstance(task, ProductPhysicalTask)]
     return tasks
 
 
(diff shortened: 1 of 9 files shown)


ludwigschwardt:
`tasks.PCLogicalTask`? :-)

bmerry:
> `tasks.PCLogicalTask`? slightly_smiling_face Goes well with "PC Load Letter".

You joke, but it's not the worst idea. Or maybe `tasks.ProductLogicalTask`, since the task is not part of the controller but the product. It is a bit of a mouthful though.

ludwigschwardt:
I put some thought into that, and only saw the joke later... 😁 I like your suggestion too.

richarms:
I agree that it would be better to have a name for `task.LogicalTask` that is distinct from `scheduler.LogicalTask`. It may mean more renaming, but one option would be to rename the parent class to `scheduler.abstractLogicalTask`, which would still be distinct from the abstract base class `scheduler.LogicalNode` in the hierarchy. However, that said, I'm fine with it as it is.

richarms (approved):

bmerry:
> I agree that it would be better to have a name for `task.LogicalTask` that is distinct from `scheduler.LogicalTask`. It may mean more renaming, but one option would be to rename the parent class to `scheduler.abstractLogicalTask`, which would still be distinct from the abstract base class `scheduler.LogicalNode` in the hierarchy. However, that said, I'm fine with it as it is.

I wasn't so keen on scheduler.AbstractLogicalTask, because it's not an abstract class i.e. it can be instantiated and used as is. I went with tasks.ProductLogicalTask etc. It's longer but there were only 3 places where I had to wrap lines to keep flake8 happy with line lengths.

ludwigschwardt:
Sorry for the late remark... There is/was also the option of leaving it as
```python
import tasks
isinstance(task, tasks.LogicalTask)
```
Although I guess `tasks` is not very descriptive either (but shorter).

james-smith-za (approved):
LGTM.

richarms (approved):

PR #592, opened 2022-05-10
NGC 544 sync cbf shutdown

bmerry:
- Strip SDP prefix from SubarrayProduct, Resources
- Remove SDP prefix from many classes
- Wait for CBF product controllers to die


bmerry:
Oops, I see this is built on top of another PR. I'll close for now and re-open once the previous change has been merged.

PR #595, opened 2022-05-16
Simulate delay sensors in interface mode

bmerry:
It's not a perfect simulation because they update immediately to the new
value rather than simulating an update at a particular time.

Closes NGC-581.

### DIFF
--- a/katsdpcontroller/fake_servers.py
+++ b/katsdpcontroller/fake_servers.py
@@ -5,7 +5,7 @@ import numbers
 from typing import Dict, Optional, Tuple
 
 import numpy as np
-from aiokatcp import Sensor, FailReply
+from aiokatcp import Sensor, Timestamp, FailReply
 from .tasks import FakeDeviceServer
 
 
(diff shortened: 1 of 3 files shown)


katsdpcontroller/fake_servers.py
```diff
@@ -48,6 +51,20 @@ def __init__(self, *args, **kwargs) -> None:
                 )
             )
 
+    async def request_delays(self, ctx, start_time: Timestamp, *delays: str) -> None:
+        """Add a new first-order polynomial to the delay and fringe correction model."""
+        # The real server only updates once the new model has gone into effect,
+        # but since we're not simulating the data path we'll just update the
+        # sensors immediately.
```
james-smith-za, line 58: This shouldn't be too hard to simulate. Maybe just something like `await asyncio.sleep(start_time - time.time())`?

Unless you specifically want to *not* simulate the data path. It would possibly break the unit tests for instance, because `time.time()` wouldn't give you repeatable output.

james-smith-za (approved):
One tiny comment / suggestion, otherwise it looks good.

katsdpcontroller/fake_servers.py
bmerry: You don't want to just block the request until the delay time kicks in. You need to schedule something in the background to change the sensor at the appropriate time, and cancel it if an earlier delay is loaded or the server is shut down, and keep track of multiple of these things. Something can certainly be done, but I would wait until someone actually has a need for it.

Unit tests also get more complicated: currently the tests mock time.time to always return something like 123456789.5, so to test the passage of time it would be necessary to do more complicated mocking.

james-smith-za (approved):

PR #603, opened 2022-05-26
Implement sensors to indicate CBF resources

bmerry:
This implements the new sensors in the draft ICD [here](https://docs.google.com/document/d/1ciRknqWsx3hmo7RrzzueZgvoTPYEQD_P5tF2iXOwvQ4/edit#).

Closes NGC-584.

### DIFF
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -35,7 +35,8 @@ import katsdpservices
 import katsdpcontroller
 from . import singularity, product_config, product_controller, scheduler, sensor_proxy
 from .defaults import LOCALHOST
-from .schemas import ZK_STATE       # type: ignore
+from .scheduler import decode_json_base64
+from .schemas import ZK_STATE, SUBSYSTEMS       # type: ignore
 from .controller import (time_request, load_json_dict, log_task_exceptions, device_server_sockname,
                          add_shared_options, extract_shared_options, make_image_resolver_factory,
                          ProductState, DeviceStatus, device_status_to_sensor_status)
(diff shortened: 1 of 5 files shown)


src/katsdpcontroller/master_controller.py
```diff
@@ -1013,6 +1029,84 @@ def mass_inform(self, name: str, *args: Any) -> None:
...
+            self.sensors['cbf-resources-maintenance'].value = cbf_resources_maintenance
+            self.sensors['cbf-resources-free'].value = cbf_resources_free
+
+    async def _update_resource_sensors_repeat(self) -> None:
+        zk = aiozk.ZKClient(self._args.zk, chroot='mesos', allow_read_only=True)
+        await zk.start()
+        try:
+            while True:
+                await self._update_resource_sensors(zk)
+                await asyncio.sleep(5)
```
james-smith-za, line 1106: Not that it's particularly important, and I don't know which behaviour will ultimately be preferable, but this will update every 5 seconds plus however long the update takes.
Alternatively, `gather()`ing these two will bring the update period closer to five seconds properly.
Not that it really matters, I don't think there's a spec, and perhaps ensuring that fully 5 seconds elapse between queries will make things easier for Mesos to breathe.

james-smith-za (approved):

src/katsdpcontroller/master_controller.py
bmerry: Yeah, I was on the fence, but I figured that if Mesos is somehow overloaded and this process is taking a significant amount of time, then it's probably reasonable to poll less frequently.

PR #610, opened 2022-07-21
Implement capture-list for CBF

bmerry:
To implement it, each PhysicalMulticast object has an enum to indicate
whether it is up, down, or unknown. Unknown streams won't appear in
capture-list.

It only lists gpucbf streams, because reporting the up/down status
requires tracking whether streams are up or down, and I didn't feel like
trying to do that for all the assorted SDP streams. It's not part of the
SDP ICD, so there is no need to do that.

Closes NGC-716.

### DIFF
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -1233,7 +1233,7 @@ class SubarrayProduct:
                 node.transmit_state = TransmitState.UP if start else TransmitState.DOWN
 
     def capture_list(self) -> Sequence[generator.PhysicalMulticast]:
-        """Return all :class:`.PhysicalMulticast` nodes that has known stream state."""
+        """Return all :class:`.PhysicalMulticast` nodes that have known stream state."""
         return [
             node for node in self.physical_graph
             if (isinstance(node, generator.PhysicalMulticast)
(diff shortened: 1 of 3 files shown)


src/katsdpcontroller/product_controller.py
```diff
@@ -1225,6 +1227,18 @@ async def capture_start_stop(self, stream_name: str, *, start: bool) -> None:
             self.find_nodes(task_type="xb", streams=[stream]),
             itertools.repeat((command,))
         )
+        for node in self.physical_graph.nodes:
+            if (isinstance(node, generator.PhysicalMulticast)
+                    and node.logical_node.name == "multicast." + stream_name):
+                node.transmit_state = TransmitState.UP if start else TransmitState.DOWN
+
+    def capture_list(self) -> Sequence[generator.PhysicalMulticast]:
+        """Return all :class:`.PhysicalMulticast` nodes that has known stream state."""
```
james-smith-za, line 1236:
```suggestion
        """Return all :class:`.PhysicalMulticast` nodes that have known stream state."""
```
?

james-smith-za (approved):
Looks good to me. One possible typo that I spotted?

bmerry:
> One possible typo that I spotted?

Fixed, thanks.

PR #613, opened 2022-09-22
Re- normalise gpu name

richarms:
After Nvidia driver updates to >11.4, the device name reported by the driver newly contains an extra `NVIDIA ` string. This will remove it, for backwards compatibility


bmerry:
The code looks fine, but the comment above should be updated to explain why it is being done.

PR #614, opened 2022-10-12
Expose dig-clip-cnt sensor under the input label

bmerry:
See NGC-796.


amishatishpatel (approved):
Nice one. Please and thank you.

PR #616, opened 2022-11-14
Pass --sync-epoch to xbgpu

bmerry:
This will need to merged synchronously with an upcoming xbgpu change to make it accept the argument.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -561,7 +561,8 @@ def _make_fgpu(
 def _make_xbgpu(
         g: networkx.MultiDiGraph,
         configuration: Configuration,
-        stream: product_config.GpucbfBaselineCorrelationProductsStream) -> scheduler.LogicalNode:
+        stream: product_config.GpucbfBaselineCorrelationProductsStream,
+        sync_time: int) -> scheduler.LogicalNode:
     ibv = not configuration.options.develop
     acv = stream.antenna_channelised_voltage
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -561,7 +561,8 @@ def _make_fgpu(
 def _make_xbgpu(
         g: networkx.MultiDiGraph,
         configuration: Configuration,
-        stream: product_config.GpucbfBaselineCorrelationProductsStream) -> scheduler.LogicalNode:
+        stream: product_config.GpucbfBaselineCorrelationProductsStream,
+        sync_time: int) -> scheduler.LogicalNode:
```
amishatishpatel, line 565: The argument added over on `xbgpu/main.py`'s parser is of `type=float`. Worth asking if this could cause some kind of conflict?

I know it's cast to string further down, then (re?)parsed anyway when `xbgpu` is invoked, but thought I'd ask anyway.

bmerry: ints are a subset of floats, so it'll be fine. This confusion probably all stems from some historical ICD that decided that the sync time had to be aligned to seconds and that's caused people to write code that bakes in that assumption even though there is nothing particularly special about moments in time that are a whole number of an arbitrary unit (second) after an arbitrary event (UNIX epoch) in an arbitrary timescale (UTC).

amishatishpatel (approved):

PR #618, opened 2022-11-15
Add an aggregate sensor for X-engine clip count

bmerry:
See NGC-797.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -8,7 +8,19 @@ import copy
 import urllib.parse
 import os.path
 from typing import (
-    List, Dict, Tuple, Set, Sequence, Iterable, Type, Union, Optional, Any, TYPE_CHECKING, cast
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Set,
+    Sequence,
+    Optional,
+    Tuple,
+    TYPE_CHECKING,
+    Type,
+    TypeVar,
+    Union,
+    cast
 )
 
 import addict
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -141,6 +156,56 @@ async def resolve(self, resolver, graph):
         self._endpoint_sensor.value = str(Endpoint(self.host, self.ports['spead']))
 
 
+class SumSensor(SimpleAggregateSensor[int]):
+    """Aggregate which takes the sum of its children.
+
+    It also tracks how many child reading are present, and sets the state to
```
james-smith-za, line 162:
```suggestion
    It also tracks how many child readings are present, and sets the state to
```

james-smith-za (changes requested):
One nit-pick, looks like a typo.
Feel free to merge after fixing.

amishatishpatel (approved):
Aside from the potential typo (that James pointed out), a more general question:
* I was under the (largely unfounded) impression that the `static_sensors` 'had to be static', and
* I was planning on creating a separate `dynamic_sensors`  to be tallied up at the `SubarrayProduct` level (during `_configure`).

Thoughts/comments welcomed, as always.

amishatishpatel:
Also, should we be expecting a corresponding PR to katsdpservices?

bmerry:
> I was under the (largely unfounded) impression that the static_sensors 'had to be static', and

That was the original intention. I had intended to rename it since the use has changed. Let me do that.

amishatishpatel (approved):
Excellent.

PR #621, opened 2022-11-18
Add fake versions of *-clip-cnt sensors

bmerry:
Also added 'synchronised' sensor for xbgpu.

Closes NGC-797.


bmerry:
Added both of you as a reminder that these sensors exist, but I feel like it only needs one review since it's copy-and-paste boilerplate.

amishatishpatel (approved):

PR #622, opened 2022-11-18
Fix up pytest deprecation warnings

bmerry:
pytest now warns that in future it's going to remove nose compatibility support for a method called `setup`. It suggests renaming it to `setup_method`. In some cases I've done that, but where it's reasonably simple I've also converted to using proper pytest fixtures.

### DIFF
--- a/test/test_product_config.py
+++ b/test/test_product_config.py
@@ -1324,356 +1324,363 @@ class TestSpectralImageStream:
         assert spec.parameters == {}
 
 
-class Fixture:
-    """Base class providing some sample config dicts"""
+@pytest.fixture
+def config() -> Dict[str, Any]:
+    return {
+        "version": "3.1",
+        "inputs": {
+            "camdata": {
+                "type": "cam.http",
+                "url": "http://10.8.67.235/api/client/1"
+            },
+            "i0_antenna_channelised_voltage": {
+                "type": "cbf.antenna_channelised_voltage",
+                "url": "spead://239.2.1.150+15:7148",
+                "antennas": ["m000", "m001", "s0003", "another_antenna"],
+                "instrument_dev_name": "i0"
+            },
+            "i0_baseline_correlation_products": {
+                "type": "cbf.baseline_correlation_products",
+                "url": "spead://239.9.3.1+15:7148",
+                "src_streams": ["i0_antenna_channelised_voltage"],
+                "instrument_dev_name": "i0"
+            },
+            "i0_tied_array_channelised_voltage_0x": {
+                "type": "cbf.tied_array_channelised_voltage",
+                "url": "spead://239.9.3.30+15:7148",
+                "src_streams": ["i0_antenna_channelised_voltage"],
+                "instrument_dev_name": "i0"
+            },
+            "i0_tied_array_channelised_voltage_0y": {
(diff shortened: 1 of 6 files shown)


test/test_product_config.py
```diff
@@ -970,51 +1090,56 @@ def make_vis(
     )
 
 
-def make_vis_4ant() -> product_config.VisStream:
+def make_vis_4ant() -> VisStream:
```
ludwigschwardt, line 1093: Is this a coincidental change, since `VisStream` is in the namespace anyway?

test/test_product_config.py
```diff
@@ -1198,7 +1327,7 @@ def test_defaults(self) -> None:
 class Fixture:
     """Base class providing some sample config dicts"""
 
-    def setup(self):
```
ludwigschwardt, line 1201: Out of interest, why keep `setup_method` here? I had issues with going to fixtures in these situations in katdal because some test helper functions needed both the fixture and a non-fixture parameter to run (can one even mix those?).

ludwigschwardt (approved):
Thanks fot this, very insightful! I'm going through the exact same exercise in katdal. So far I've just renamed setup -> setup_method but this inspires me to try harder to get fixtures in there.

ludwigschwardt:
I also got my warnings down from 269928 (!! mostly np.bool) to 1 -- the same one about the `imp` module. I'm also looking forward to katpoint 1.0!

test/test_product_config.py
bmerry: Yes - happened to spot it and it bugged me.

test/test_product_config.py
bmerry: Good point. I've now turned each of the dictionaries in Fixture into its own pytest fixture.

Accessing fixtures from helper functions is one thing I find painful in pytest, particularly compared to the old unittest/nose style. I usually end up having to pass them into the helper function. The alternative approach is to make the helper a fixture itself (a `pytest.fixture`-decorated function that returns a function), which does let you hide the dependent fixtures, but then you have to declare the helper as a fixture parameter on every test that uses it.

test/test_product_config.py
```diff
@@ -1710,55 +1717,55 @@ async def test_sim(self) -> None:
...
-        self.config_sim['outputs']['bcp']['src_streams'] = ['vis']
+    async def test_cyclic(self, config_sim: Dict[str, Any]) -> None:
+        config_sim['outputs']['bcp']['src_streams'] = ['vis']
         with pytest.raises(ValueError):
-            await Configuration.from_config(self.config_sim)
+            await Configuration.from_config(config_sim)
 
-    async def test_bad_stream(self) -> None:
-        self.config['outputs']['l0']['continuum_factor'] = 3
+    async def test_bad_stream(self, config: Dict[str, Any]) -> None:
```
ludwigschwardt, line 1725: How do you know if it picks `TestConfiguration.config` or the module-level `config`? Pytest magic?

ludwigschwardt (approved):

test/test_product_config.py
bmerry: pytest has similar scoping rules to Python. Class takes precedence over the module (and the module over conftest.py).

PR #625, opened 2022-11-23
Add xbgpu aggregate sync sensor

amishatishpatel:
Considering all X-engines should have their `synchronised` sensors
go to `False` when an F-engine goes missing, perhaps I should change
the calculation of the aggregated `synchronised` to ensure all are `False`?
(To be more explicit from a readability POV.)

Either way, I've run through the various ways of correlator degradation to
ensure the status does move through the values accordingly:
* SyncSensor goes to `ERROR` when X-engines are showing `False`,
* ~~SyncSensor goes to `FAILURE` when an X-engine disappears.~~ As per the discussion below, this case was removed.

Resolves: NGC-665.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -209,11 +209,15 @@ class SumSensor(SimpleAggregateSensor[int]):
 
 
 class SyncSensor(SimpleAggregateSensor[bool]):
-    """Aggregate which takes the logical AND of its children.
+    """Aggregate which tracks if its children are synchronised.
 
-    The status computes as ERROR if any of the child readings are False
-    or absent.
-    TODO: Update the comment to indicate the updated logic.
+    In this case,
+    - a 'synchronised' child has a reading of (True, NOMINAL), and
+    - an 'unsynchronised' child has a reading of (False, ERROR).
+
+    This aggregate sensor does not make provision for a FAILURE status
+    as this functionality is considered covered in a dedicated Device
+    Status sensor.
     """
 
     def __init__(
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -216,7 +216,7 @@ class SyncSensor(SimpleAggregateSensor[bool]):
     """
 
     def __init__(
-        self, target: SensorSet, sensor_type: Type[_T],
+        self, target: SensorSet, sensor_type: Type[bool],
         name: str, description: str, units: str = "",
         *,
         auto_strategy: Optional["SensorSampler.Strategy"] = None,
@@ -228,7 +228,7 @@ class SyncSensor(SimpleAggregateSensor[bool]):
         self.name_regex = name_regex
         self.children = children
         self._known = 0
-        self.synchronised = None
+        self.synchronised = False
         self.curr_values: List[bool] = []
 
         super().__init__(
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -211,8 +211,8 @@ class SumSensor(SimpleAggregateSensor[int]):
 class SyncSensor(SimpleAggregateSensor[bool]):
     """Aggregate which takes the logical AND of its children.
 
-    The status computes as ERROR if any of the child readings are False, and
-    FAILURE if they aren't all present.
+    The status computes as ERROR if any of the child readings are False
+    or absent.
     """
 
     def __init__(
@@ -228,8 +228,7 @@ class SyncSensor(SimpleAggregateSensor[bool]):
         self.name_regex = name_regex
         self.children = children
         self._known = 0
-        self.synchronised = False
-        self.curr_values: Dict[str, bool] = {}
+        self._total_in_sync = 0
 
         super().__init__(
             target, sensor_type, name, description, units,
@@ -243,7 +242,8 @@ class SyncSensor(SimpleAggregateSensor[bool]):
     def aggregate_add(self, sensor: Sensor[_T], reading: Reading[_T]) -> bool:
         assert isinstance(reading.value, bool)
         if reading.status.valid_value():
-            self.curr_values[sensor.name] = reading.value
+            if reading.value is True:
+                self._total_in_sync += 1
             self._known += 1
             return True
         return False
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -208,6 +208,63 @@ def aggregate_compute(self) -> Tuple[Sensor.Status, int]:
...
+            auto_strategy_parameters=auto_strategy_parameters
+        )
+
+    def filter_aggregate(self, sensor: Sensor) -> bool:
+        return bool(self.name_regex.fullmatch(sensor.name))
+
+    def aggregate_add(self, sensor: Sensor[_T], reading: Reading[_T]) -> bool:
+        assert isinstance(reading.value, bool)
+        if reading.status.valid_value():
+            self.curr_values[sensor.name] = reading.value
```
bmerry, line 246: Not sure why you need to keep a dictionary by sensor. All you need is a count of how many are synchronised, which `aggregate_compute` will compare to `children`.

src/katsdpcontroller/generator.py
```diff
@@ -208,6 +208,63 @@ def aggregate_compute(self) -> Tuple[Sensor.Status, int]:
         return (status, self._total)
 
 
+class SyncSensor(SimpleAggregateSensor[bool]):
+    """Aggregate which takes the logical AND of its children.
+
+    The status computes as ERROR if any of the child readings are False, and
+    FAILURE if they aren't all present.
```
bmerry, line 215: My gut feeling is that if an X-engine vanishes, it's safe to say that it's not synchronised. We are thus sure that the instrument is not synchronised (for which ERROR is appropriate) rather than being forced to say that we don't know.

(technically there are some failure modes where we lose comms to the X-engine but it's still running fine, but I think in such cases we should still say that it's not synchronised, just from a health-check point of view)

src/katsdpcontroller/generator.py
```diff
@@ -695,7 +752,12 @@ def get_baseline_index(a1, a2):
         SumSensor(sensors, int, f"{stream.name}-xeng-clip-cnt",
                   "Number of visibilities that saturated",
                   name_regex=re.compile(rf"xb\.{stream.name}\.[0-9]+\.xeng-clip-cnt"),
-                  children=stream.n_substreams)
+                  children=stream.n_substreams),
+        SyncSensor(sensors, bool, f"{stream.name}-xengs-synchronised",
+                   "For the latest accumulation, was data present from all F-Engines \
+                   for all X-Engines",
+                   name_regex=re.compile(rf"xb\.{stream.name}\.[0-9]+\.synchronised"),
```
bmerry, line 759: Just realised that sense we're pasting a user-provided string into a regex, it probably ought to be escaped (that also applies to the existing SumSensor). Forget the exact syntax, but something like
```suggestion
                   name_regex=re.compile(rf"xb\.{re.escape(stream.name)}\.[0-9]+\.synchronised"),
```
There are limitations on the stream name character set that mean it probably doesn't matter, but better safe than sorry.

bmerry (changes requested):

bmerry:
> Considering all X-engines should have their synchronised sensors
go to False when an F-engine goes missing, perhaps I should change
the calculation of the aggregated synchronised to ensure all are False?
(To be more explicit from a readability POV.)

No, there are failure modes (e.g. bad network cables in a Clos network) that would cause only some X-engines to lose data. Also just dropped packets will do it.

bmerry (changes requested):
The design looks right now. Just nitpicks.

src/katsdpcontroller/generator.py
```diff
@@ -208,6 +208,63 @@ def aggregate_compute(self) -> Tuple[Sensor.Status, int]:
...
+class SyncSensor(SimpleAggregateSensor[bool]):
+    """Aggregate which tracks if its children are synchronised.
+
+    In this case,
+    - a 'synchronised' child has a reading of (True, NOMINAL), and
+    - an 'unsynchronised' child has a reading of (False, ERROR).
+
+    This aggregate sensor does not make provision for a FAILURE status
+    as this functionality is considered covered in a dedicated Device
+    Status sensor.
```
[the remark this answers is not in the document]
amishatishpatel: I was trying to word something like "the SyncSensor doesn't have logic to move to FAILURE as that's the duty of a dedicated device-health sensor". Kind of like what you described in your [first comment](https://github.com/ska-sa/katsdpcontroller/pull/625#discussion_r1030420953)
* i.e. We may not be able to contact it, but the device is not **uncontactable**.

I think I went through a PR of yours recently (Dig power?) that discusses the "duty to reset the average power count is on the user, not the logic".
* Figured it's a good idea to be explicit about what this sensor does and doesn't do.
* But maybe I was over-reaching with this docstring description...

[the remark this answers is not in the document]
james-smith-za: I don't feel that the paragraph adds much value. I'd suggest just removing it.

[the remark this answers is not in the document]
amishatishpatel: Yeah fair point, shall do.

src/katsdpcontroller/generator.py
```diff
@@ -208,6 +208,59 @@ def aggregate_compute(self) -> Tuple[Sensor.Status, int]:
...
+    def __init__(
+        self, target: SensorSet, sensor_type: Type[bool],
+        name: str, description: str, units: str = "",
+        *,
+        auto_strategy: Optional["SensorSampler.Strategy"] = None,
+        auto_strategy_parameters: Iterable[Any] = (),
+        name_regex: re.Pattern,
+        children: int
+    ) -> None:
+        self.target = target
```
james-smith-za, line 228: This line isn't necessary, just pass `target` through to the `super()` method. It does the assignment (which may even just clobber this one? I'm not too sure.)

src/katsdpcontroller/generator.py
```diff
@@ -208,6 +208,59 @@ def aggregate_compute(self) -> Tuple[Sensor.Status, int]:
...
+        self.name_regex = name_regex
+        self.children = children
```
james-smith-za, line 230: I'd just suggest making this `n_children` just so that it's obvious that it's a count, rather than a set or list or something.
Same goes for the parameter above.

bmerry (approved):

src/katsdpcontroller/generator.py
```diff
@@ -208,6 +208,59 @@ def aggregate_compute(self) -> Tuple[Sensor.Status, int]:
...
+class SyncSensor(SimpleAggregateSensor[bool]):
+    """Aggregate which tracks whether its children are synchronised.
+
+    In this case,
+    - a 'synchronised' child has a reading of (True, NOMINAL), and
+    - an 'unsynchronised' child has a reading of (False, ERROR).
+    """
+
+    def __init__(
+        self, target: SensorSet, sensor_type: Type[bool],
```
james-smith-za, line 220: I'd even suggest removing the type from this class, and just hard-coding it to `bool` when you call `super().__init__()` lower down.

It's unlikely that the user would need to change this.

I'm tempted to suggest renaming this to something like `AllSensor` or something to do with its boolean-ness, because it could well be reused for other things that aren't sync statuses.

But looking at the ICD, there don't seem to be other potential candidates so perhaps it's best not to worry about that at this stage.

james-smith-za (changes requested):
I'm also happy with the overall structure of this, just minor nit-picks in my case.

src/katsdpcontroller/generator.py
amishatishpatel: Yeah I'm not sure why I did that. Maybe it was initially a reminder to myself regarding what I have access to. Updating accordingly.

src/katsdpcontroller/generator.py
amishatishpatel: @bmerry permission to update the corresponding `sensor_type` arg in `SumSensor` to `int` alongside this?

bmerry: Yes, definitely worth making things consistent. I think I'd originally thought SumSensor should be generic (so that we could accumulate `float` as well), but floating-point rounding errors don't play nicely with the implementation (continually adding things in then subtracting them out later).

PR #627, opened 2022-11-28
Add an assertion to keep mypy happy

bmerry:
Somehow this slipped into master without being picked up.


james-smith-za (approved):
Not sure why Jenkins still says "failed" even after fixing the issue.

bmerry:
> Not sure why Jenkins still says "failed" even after fixing the issue.

Somehow I'd missed an instance, and pre-commit didn't call me on it for some reason. Hopefully fixed now.

james-smith-za (approved):

PR #630, opened 2022-12-01
Add fake fgpu rx.device-status sensor

james-smith-za:
I find myself wondering whether the DeviceStatus enum is useful enough to actually be moved into aiokatcp instead.

Closes NGC-847.

### DIFF
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -2,6 +2,7 @@
 
 import json
 import numbers
+from enum import Enum
 from typing import Dict, Optional, Tuple
 
 import numpy as np
@@ -15,6 +16,14 @@ def _format_complex(value: numbers.Complex) -> str:
     This is copied from katgpucbf.
     """
     return f"{value.real}{value.imag:+}j"
+
+
+class DeviceStatus(Enum):
+    """Discrete `device-status` readings."""
+
+    OK = 1
+    DEGRADED = 2
+    FAIL = 3
 
 
 class FakeFgpuDeviceServer(FakeDeviceServer):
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/fake_servers.py
```diff
@@ -108,6 +117,16 @@ def __init__(self, *args, **kwargs) -> None:
...
             )
 
+        self.sensors.add(
+            Sensor(
+                DeviceStatus,
+                "rx.device-status",
+                "The F-engine is receiving a good, clean digitiser stream",
+                default=DeviceStatus.DEGRADED,
+                initial_status=Sensor.Status.WARN,
+            )
```
amishatishpatel, line 127: I was having a general thought about "why these values should be what they are", and I see that the ICD doesn't actually mention default value/status combinations for sensors. Do we know why that is?

amishatishpatel (commented):
Just a comment before I explicitly approve. Admittedly probably outside the scope of the PR.

Regarding `DeviceStatus`, what about moving it inside `DeviceStatusSensor` itself? (like aiokatcp's Status and Sensor)

src/katsdpcontroller/fake_servers.py
james-smith-za: Do you mean why the ICD doesn't mention default values? Well, what the sensors show should reflect what is actually going on. This is just a "fake" server so it's got to have some values to work with.

In this case, the device-status sensor is supposed to be a function of what is underneath it, and since there are some sensors that are in `ERROR`, I made the status of this correspond to what it would in a real case.

amishatishpatel: > Do you mean why the ICD doesn't mention default values?

Indeed indeed. Although, looking at fgpu's [populate_sensors](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/fgpu/engine.py#L723-L786), I see the foundation of my point crumbling as
* there doesn't seem to be any consistency on whether a sensor needs any combination of `default` and `initial_status` (upon instantiation).

Not to say that it *should* require either/both of those.

> In this case, the device-status sensor is supposed to be a function of what is underneath it, and since there are some sensors that are in `ERROR`, I made the status of this correspond to what it would in a real case.

Indeed indeed once again. I follow your logic (and also checked why these values were so during my review) - no qualms there.

bmerry (changes requested):
LGTM apart from one stylistic niggle (I don't feel a need to re-review after fixing that). I'd be happy to move DeviceStatus to aiokatcp (probably in core.py).

PR #631, opened 2022-12-06
Add hardcoded ddc-mix-freq sensor

james-smith-za:
For the time being, this just reads zero, to fulfil the obligations that we have as per the ICD (NGC-831).

I think ideally, the value should be read from the stream configuration, and I did initially try poking at that a bit, but I don't understand enough about how this works to be able to be sure I haven't borken something.

So perhaps the proper logic should be introduced as part of NGC-564 or NGC-565.


src/katsdpcontroller/generator.py
```diff
@@ -762,6 +762,9 @@ def get_baseline_index(a1, a2):
         Sensor(int, f"{stream.name}-n-chans-per-substream",
                "Number of channels in each substream for this x-engine stream",
                default=stream.n_chans_per_substream, initial_status=Sensor.Status.NOMINAL),
+        Sensor(float, f"{stream.name}.ddc-mix-freq",
+               "F-engine DDC mixer frequency, where used. 0 where n/a",
+               default=0, initial_status=Sensor.Status.NOMINAL),
```
amishatishpatel, line 767: Might request you add the `units` kwarg, like in `fake_sever.py`'s `dig-pwr-dbfs` (or katgpucbf's `dsim/server.py` for `adc-sample-rate`).

amishatishpatel (changes requested):

src/katsdpcontroller/generator.py
james-smith-za: That's a good point.

amishatishpatel: Out of interest, any reason we went with `mix-freq` instead of `mixer-freq`?

src/katsdpcontroller/generator.py
```diff
@@ -764,7 +764,7 @@ def get_baseline_index(a1, a2):
                default=stream.n_chans_per_substream, initial_status=Sensor.Status.NOMINAL),
         Sensor(float, f"{stream.name}.ddc-mix-freq",
                "F-engine DDC mixer frequency, where used. 0 where n/a",
-               default=0, initial_status=Sensor.Status.NOMINAL),
+               units="Hz", default=0.0, initial_status=Sensor.Status.NOMINAL),
```
amishatishpatel, line 767: Ah, yes. I was actually looking around to see if the `sensor_type` arg was used, forgot to comment.

amishatishpatel (approved):
Happy with the update. No need to necessarily change from `mix-freq` to `mixer-freq` (unless Bruce reckons otherwise?).

src/katsdpcontroller/generator.py
james-smith-za: That's the very first one.

src/katsdpcontroller/generator.py
james-smith-za: Uh, now you're asking the hard questions. I just went with what's written in the ICD, I think it's what was used previously.

src/katsdpcontroller/generator.py
amishatishpatel: Argh, apologies. I always miss the first arg for some reason.

src/katsdpcontroller/generator.py
bmerry: The original ICD seemed to name things like they were being charged by the letter. So bls, chans, ants, mix, pwr instead of baselines, channels, antennas, mixer, power.

bmerry (commented):
LGTM, although maybe we should at least keep the code consistent about dots versus dashes and then make a wholesale change, rather than introducing a mix?

james-smith-za:
> LGTM, although maybe we should at least keep the code consistent about dots versus dashes and then make a wholesale change, rather than introducing a mix?

I did sort of think of that but my main reservation was, it felt wrong to deliberately do something "wrong" that I'm going to have to change again later.

But the argument for consistency is a valid one.

PR #632, opened 2022-12-06
Add fake xengine chan-range sensor

james-smith-za:
Closes NGC-837.

I did a bit of refactoring as well while I was here, didn't think anyone would mind.

### DIFF
--- a/src/katsdpcontroller/tasks.py
+++ b/src/katsdpcontroller/tasks.py
@@ -10,11 +10,13 @@ from contextlib import asynccontextmanager
 from typing import (
     AsyncContextManager,
     AsyncGenerator,
+    Callable,
     List,
     MutableMapping,
     Optional,
     Set,
     Type,
+    TypeVar,
     Union
 )
 
@@ -30,6 +32,7 @@ from . import scheduler, sensor_proxy, product_config
 from .consul import ConsulService, CONSUL_PORT
 from .defaults import LOCALHOST
 
+_T = TypeVar("_T")
 
 logger = logging.getLogger(__name__)
 # Name of edge attribute, as a constant to better catch typos
@@ -735,6 +738,11 @@ class FakeDeviceServer(aiokatcp.DeviceServer):
         """Respond to any unknown requests with an empty reply."""
         ctx.reply(aiokatcp.core.Message.OK)
         await ctx.drain()
+
+    def get_command_value(self, value_type: Callable[..., _T], parameter_string: str) -> _T:
+        """Return the value passed to the fake device as a cli parameter."""
+        position = self.logical_task.command.index(parameter_string)
+        return value_type(self.logical_task.command[position + 1])
 
 
 @asynccontextmanager
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -167,6 +167,10 @@ class FakeFgpuDeviceServer(FakeDeviceServer):
 class FakeXbgpuDeviceServer(FakeDeviceServer):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
+        channel_offset_pos = self.logical_task.command.index("--channel-offset-value")
+        channel_offset = int(self.logical_task.command[channel_offset_pos + 1])
+        channels_per_substream_pos = self.logical_task.command.index("--channels-per-substream")
+        channels_per_substream = int(self.logical_task.command[channels_per_substream_pos + 1])
         self.sensors.add(
             Sensor(
                 bool,
@@ -182,6 +186,15 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                 "xeng-clip-cnt",
                 "Number of visibilities that saturated",
                 default=0,
+                initial_status=Sensor.Status.NOMINAL
+            )
+        )
+        self.sensors.add(
+            Sensor(
+                str,
+                "chan-range",
+                "The range of channels processed by this XB-engine, inclusive",
+                default=f"({channel_offset},{channel_offset + channels_per_substream - 1})",
                 initial_status=Sensor.Status.NOMINAL
             )
         )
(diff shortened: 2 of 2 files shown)


src/katsdpcontroller/tasks.py
```diff
@@ -736,6 +739,11 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
         ctx.reply(aiokatcp.core.Message.OK)
         await ctx.drain()
 
+    def get_command_value(self, value_type: Callable[..., _T], parameter_string: str) -> _T:
+        """Return the value passed to the fake device as a cli parameter."""
```
bmerry, line 743:
```suggestion
        """Return the value passed to the fake device as a CLI parameter."""
```

src/katsdpcontroller/tasks.py
```diff
@@ -736,6 +739,11 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
         ctx.reply(aiokatcp.core.Message.OK)
         await ctx.drain()
 
+    def get_command_value(self, value_type: Callable[..., _T], parameter_string: str) -> _T:
```
bmerry, line 742: I'd remove "string" from the name - we don't generally include the type of an argument in its name. And I might rename "parameter" to "argument" (since that's what Python calls command-line arguments e.g. in `argparse`). Then while I'm bikeshedding, what do you think about calling the method `get_command_argument`?

src/katsdpcontroller/tasks.py
```diff
@@ -736,6 +739,11 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
...
+        """Return the value passed to the fake device as a cli parameter."""
+        position = self.logical_task.command.index(parameter_string)
+        return value_type(self.logical_task.command[position + 1])
```
bmerry, line 745: Does taking `value_type` actually add much? I don't think `get_command_value(int, "--foo")` is any more convenient than `int(get_command_value("--foo"))`. I guess it does ensure that you don't forget to convert at all.

src/katsdpcontroller/tasks.py
bmerry, line 742: The implementation is quite fragile e.g. it won't work if one uses `--foo=bar` syntax, and it'll crash if the argument is not found. I'm totally fine with that (it's not expected to deal with arbitrary commands), but the methodology should be documented in the docstring.

bmerry (changes requested):
I'm in favour of the refactoring. Just some bikeshedding.

src/katsdpcontroller/tasks.py
james-smith-za: Would `--foo=bar` be a single string instead of subsequent arguments in the list?

amishatishpatel (commented):
I think Bruce has pretty much covered it, just the one comment/question from me (for now).

src/katsdpcontroller/tasks.py
bmerry: That would be a single string. But as I say, I think it's easier to just not use that syntax when constructing the command line, as long as this expectation is documented.

I did think about trying to leverage `argparse` to handle the different cases more cleanly, but I think that runs into its own issues because some of the commands are wrappers that use `--` to signal that all following arguments are to be passed to the wrapped program.

bmerry: Sure, `argument_name` sounds good. I was trying to think of a way to name it that makes it obvious that it should include the leading `--`, but I can't think of anything both succinct and clear.

bmerry: BTW if you do keep `value_type`, it can probably be more strictly typed as `Callable[[str], _T]`.

src/katsdpcontroller/tasks.py
james-smith-za: So without the built-in conversion, it would be something like 

```python
def get_command_argument(self, argument_name: str) -> str:
   ...
   return self.logical_task.command[position + 1]
```

Then the caller would use it like this
```
my_int_value = int(self.get_command_argument("--my-int-value"))
```

So no `Callable[[str], [_T]]` needed basically.

I've personally got no preference as to which is better, but since I've already done the one approach it's not going to save me any typing to undo it.

src/katsdpcontroller/tasks.py
```diff
@@ -739,9 +739,16 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
         ctx.reply(aiokatcp.core.Message.OK)
         await ctx.drain()
 
-    def get_command_value(self, value_type: Callable[..., _T], parameter_string: str) -> _T:
-        """Return the value passed to the fake device as a cli parameter."""
-        position = self.logical_task.command.index(parameter_string)
+    def get_command_argument(self, value_type: Callable[[str], _T], argument_name: str) -> _T:
+        """Return the value passed to the fake device as a CLI parameter.
+
+        This function is not completely robust, it assumes that parameters are
```
bmerry, line 745:
```suggestion
        This function is not completely robust, as it assumes that parameters are
```

src/katsdpcontroller/tasks.py
```diff
@@ -739,9 +739,16 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
...
+        only passed as separate strings, such as `--foo bar`, and not using
```
bmerry, line 746:
```suggestion
        only passed as separate strings, such as ``--foo bar``, and not using
```

src/katsdpcontroller/tasks.py
```diff
@@ -739,9 +739,16 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
...
+        `--foo=bar` syntax. The argument name (e.g. `--foo`) is located in
```
bmerry, line 747:
```suggestion
        ``--foo=bar`` syntax. The argument name (e.g. ``--foo``) is located in
```

src/katsdpcontroller/tasks.py
```diff
@@ -739,9 +739,16 @@ async def unhandled_request(self, ctx, req: aiokatcp.core.Message) -> None:
...
+        the logical tasks's comamnd list, and it is assumed that the "value"
+        passed is simply the string next in the list.
```
bmerry, line 749:
```suggestion
        passed is simply the next string in the list.
```

src/katsdpcontroller/tasks.py
bmerry: I'm happy to keep it as is then.

bmerry (changes requested):

PR #633, opened 2022-12-08
Add .source and .revision sensors for tasks

bmerry:
The data for these are taken from Docker labels that Jenkins (both SDP and CBF) provides when it builds the images. To handle this, the ImageLookup framework needed to be reworked to return an object describing the image rather than just a string.

There is also some rework of the `.host` and `.version` sensors to make them behave more uniformly relative to the `.state` and `.mesos-state` sensors.

When in interface mode, these sensors will exist, but will have unknown values since there are no Docker images involved.

This does not yet provide these sensors for product controllers. That will need some rework in the master controller Zookeeper schema.


james-smith-za (approved):
Can't see any obvious problems, but I'm somewhat out of my depth with how the controller actually works.

PR #635, opened 2022-12-09
Convert mypy to a pre-commit local hook

bmerry:
This means it will be run on the installed virtualenv rather than creating it's own isolated one, and hence has access to the dependencies for type checking.

Also update mypy to the latest version, because the previous one was giving an internal error.

### DIFF
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -254,10 +254,9 @@ class ProductManagerBase(Generic[_P]):
         self._image_resolver_factory = image_resolver_factory
         self._rewrite_gui_urls = rewrite_gui_urls
 
-    @abstractmethod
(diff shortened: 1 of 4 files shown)


src/katsdpcontroller/master_controller.py
```diff
@@ -254,10 +254,9 @@ def __init__(self, args: argparse.Namespace, server: aiokatcp.DeviceServer,
         self._image_resolver_factory = image_resolver_factory
         self._rewrite_gui_urls = rewrite_gui_urls
 
-    @abstractmethod
```
bmerry, line 257: mypy was complaining that subclasses were using `super()` to call the abstract method. I think it reckons abstract methods should be pure virtual.

ludwigschwardt (approved):

PR #636, opened 2022-12-12
Add .source and .revision sensors for product controllers

bmerry:
To support this, the Zookeeper schema for persisting product controller information over MC restarts is updated to version 5, holding a JSON representation of the Image class. Support for versions 1-3 is removed to simplify the code (they're ancient and not needed since there are no running product controllers using them). If a v4 product controller is loaded when the MC starts up, it uses a somewhat hacky approach to synthesize an Image object (Image.from_path method).

Closes NGC-841.

- [x] Re-test the final modifications in the lab once the link to site is working again


james-smith-za (approved):
I don't see any obvious problems, but my general lack of knowledge of how the master controller works is my weakness here.

richarms (approved):

PR #637, opened 2022-12-12
Run pyupgrade --py38-plus

bmerry:
It's purely mechanical changes. Mostly use of f-strings, but also some
set comprehensions, generator expressions instead of list
comprehensions, new-style `super`, and removal of `object` as a base
class.


ludwigschwardt (approved):
I should go there soon too...

PR #642, opened 2022-12-13
Second round of applying black

bmerry:
To keep review sizes sane, it's applied to the test directory, and (for now) with `-S` to prevent it changing string quotes.

The second (much smaller) commit is some manual changes to improve readability, and bears more careful review.

### DIFF
--- a/test/test_product_config.py
+++ b/test/test_product_config.py
@@ -40,8 +40,12 @@ from katsdpcontroller.product_config import (
 
 from . import fake_katportalclient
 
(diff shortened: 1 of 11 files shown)


test/test_product_config.py
```diff
@@ -1,6 +1,7 @@
 """Tests for :mod:`katsdpcontroller.product_config`."""
 
 import copy
+import re
```
james-smith-za, line 4: Did black do this? There are some `re.escape()` calls down below, which strikes me as an odd thing for an automated tool to add.

james-smith-za (approved):

test/test_product_config.py
bmerry: Go read the PR description again.

PR #647, opened 2022-12-14
Add fake xbgpu.rx.device-status sensor

james-smith-za:
Closes NGC-848

### DIFF
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -257,6 +257,16 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                 initial_status=Sensor.Status.NOMINAL
             )
         )
+        self.sensors.add(
+            Sensor(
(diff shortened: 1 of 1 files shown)


james-smith-za:
Happy to wait until black stuff is done.

bmerry:
> Happy to wait until black stuff is done.

I'm doing it piecemeal partly to avoid conflicting with stuff that I know is in progress.

bmerry (changes requested):
Given that 80% of the arguments to Sensor are duplicated from the equivalent sensor in fgpu, can I suggest adding an `_add_rx_device_status_sensor` helper that takes the description as a parameter?

james-smith-za:
> Given that 80% of the arguments to Sensor are duplicated from the equivalent sensor in fgpu, can I suggest adding an `_add_rx_device_status_sensor` helper that takes the description as a parameter?

Better?

bmerry (approved):

PR #649, opened 2022-12-14
Final round of black conversion

bmerry:
This runs black on the remaining files, and also adds it to pre-commit. As before, it's split into multiple commits which require different levels of scrutiny.

### DIFF
--- a/src/katsdpcontroller/tasks.py
+++ b/src/katsdpcontroller/tasks.py
@@ -102,13 +102,11 @@ BATCH_RUNTIME_REL = Histogram(
 class CaptureBlockState(scheduler.OrderedEnum):
     """State of a single capture block."""
 
-    # fmt: off
-    INITIALISING = 0         # Only occurs briefly on construction
-    CAPTURING = 1            # capture-init called, capture-done not yet called
-    BURNDOWN = 2             # capture-done returned, but real-time processing still happening
-    POSTPROCESSING = 3       # real-time processing complete, running batch processing
-    DEAD = 4                 # fully complete
-    # fmt: on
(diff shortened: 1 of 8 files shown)


src/katsdpcontroller/tasks.py
```diff
@@ -102,11 +102,13 @@
...
-    BURNDOWN = 2  # capture-done returned, but real-time processing still happening
-    POSTPROCESSING = 3  # real-time processing complete, running batch processing
-    DEAD = 4  # fully complete
+    # fmt: off
+    INITIALISING = 0         # Only occurs briefly on construction
+    CAPTURING = 1            # capture-init called, capture-done not yet called
+    BURNDOWN = 2             # capture-done returned, but real-time processing still happening
+    POSTPROCESSING = 3       # real-time processing complete, running batch processing
+    DEAD = 4                 # fully complete
+    # fmt: on
```
james-smith-za, line 111: I mean it's nice to have the comments lined up like that, but I don't really mind how black does it. To me the `fmt` comments above and below are almost more unsightly.

I take it that this was for aesthetic reasons, and not functional ones.
As such, this is merely my own opinion, and has no real bearing on the merge at all.

james-smith-za (approved):

src/katsdpcontroller/tasks.py
bmerry: Yes, purely aesthetic. I'll revert it; there are other similar places where black ruined the alignment and I didn't get around to doing this, so it's currently not consistent.

james-smith-za (approved):

PR #650, opened 2022-12-14
Make code no-implicit-optional safe (mypy)

bmerry:
No longer need to pass --implicit-optional to mypy. These are
automated changes (made with the no-implicit-optional package).


james-smith-za (approved):
Explicit is better than implicit I guess.

PR #651, opened 2023-01-05
Use fullstops for hierarchical separation in sensor names

james-smith-za:
This change affects the product controller. Previously, hyphens were
used basically everywhere.

The CBF ICD with CAM requires that the KATCP guidelines (i.e. using a
fullstop to separate hierarchical levels), which is what prompted this
change.

Note: there are some not-exclusively-CBF sensors that are affected here.
I'm not sure whether or how much this will affect SDP.

Addresses NGC-854.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2013-2019, National Research Foundation (SARAO)
+Copyright (c) 2013-2023, National Research Foundation (SARAO)
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -146,7 +146,7 @@ class FakeFgpuDeviceServer(FakeDeviceServer):
             self.sensors.add(
                 Sensor(
                     int,
-                    f"input{pol}.rx-timestamp",
+                    f"input{pol}.rx.timestamp",
                     "The timestamp (in samples) of the last chunk of data received "
                     "from the digitiser",
                     default=-1,
@@ -156,7 +156,7 @@ class FakeFgpuDeviceServer(FakeDeviceServer):
             self.sensors.add(
                 Sensor(
                     Timestamp,
-                    f"input{pol}.rx-unixtime",
+                    f"input{pol}.rx.unixtime",
                     "The timestamp (in UNIX time) of the last chunk of data received "
                     "from the digitiser",
                     default=Timestamp(-1.0),
@@ -166,7 +166,7 @@ class FakeFgpuDeviceServer(FakeDeviceServer):
             self.sensors.add(
                 Sensor(
                     Timestamp,
-                    f"input{pol}.rx-missing-unixtime",
+                    f"input{pol}.rx.missing-unixtime",
                     "The timestamp (in UNIX time) when missing data was last detected",
                     default=Timestamp(-1.0),
                     initial_status=Sensor.Status.NOMINAL,
@@ -224,7 +224,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
         self.sensors.add(
             Sensor(
                 bool,
-                "synchronised",
+                "rx.synchronised",
                 "For the latest accumulation, was data present from all F-Engines.",
                 default=True,
                 initial_status=Sensor.Status.NOMINAL,
@@ -251,7 +251,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
         self.sensors.add(
             Sensor(
                 int,
-                "input-rx-timestamp",
+                "rx.timestamp",
                 "The timestamp (in samples) of the last chunk of data received from an F-engine",
                 default=-1,
                 initial_status=Sensor.Status.ERROR,
@@ -260,7 +260,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
         self.sensors.add(
             Sensor(
                 Timestamp,
-                "input-rx-unixtime",
+                "rx.unixtime",
                 "The timestamp (in UNIX time) of the last chunk of data received "
                 "from an F-engine",
                 default=Timestamp(-1.0),
@@ -270,7 +270,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
         self.sensors.add(
             Sensor(
                 Timestamp,
-                "input-rx-missing-unixtime",
+                "rx.missing-unixtime",
                 "The timestamp (in UNIX time) when missing data was last detected",
                 default=Timestamp(-1.0),
                 initial_status=Sensor.Status.NOMINAL,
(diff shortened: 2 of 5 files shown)


ludwigschwardt:
So this is different to the way the old correlator did it?

james-smith-za:
> So this is different to the way the old correlator did it?

This change brings it closer to the way the old correlator behaved. The old correlator was (I think) not entirely consistent in this but it mostly used fullstops to separate hierarchical levels in sensor names.

My main reservation is that some non-CBF-exclusive stuff is touched as well, so the behaviour of SDP products will be (slightly) affected by this. I'm going to have a look at it again tomorrow morning to make sure that the changes are consistent at least within katsdpcontroller, but I'm not sure what else in the ecosystem might make assumptions about these sensor names.

ludwigschwardt:
On the SDP side most (all?) sensor interactions are via telstate, which normalises all punctuation to underscores. So it's probably OK.

bmerry (approved):
LGTM. I'm not sure exactly what you meant by

> Note: there are some not-exclusively-CBF sensors that are affected here.
I'm not sure whether or how much this will affect SDP.

I assume this refers to the `.destination` sensors. Those were added specifically for CBF ([here](https://github.com/ska-sa/katsdpcontroller/commit/b44edcc2f8b6a642b58f5cd41c37078302329f3c)) so I don't think anything should be depending on them existing for SDP streams.

bmerry (changes requested):
Actually, not approved, because you need to update the unit tests to match. Currently there are lots of failures.

bmerry:
> So this is different to the way the old correlator did it?

It's different to the old ICD, which separated the stream name, the input label and the individual sensor with dashes. I can't comment on whether that matches the old implementation.

amishatishpatel:
Before I request a re-review, I wanted to clarify whether `_make_xbgpu` needed a `sensor_renames` for making the Rx-timestamp sensors part of the stream.
* It doesn't *seem* necessary to me,
    * `baseline_correlation_products` doesn't really have an `{input_labels}` identifier (like fgpu does/uses),
    * So, in my mind, the name would look vastly similar to the `xb.{stream.name}.{i}` value that exists.

Maybe I'm not seeing the reasoning correctly. Any thoughts are welcomed indeed.

bmerry:
> Before I request a re-review, I wanted to clarify whether `_make_xbgpu` needed a `sensor_renames` for making the Rx-timestamp sensors part of the stream.
> 
>     * It doesn't _seem_ necessary to me,
>       
>       * `baseline_correlation_products` doesn't really have an `{input_labels}` identifier (like fgpu does/uses),
>       * So, in my mind, the name would look vastly similar to the `xb.{stream.name}.{i}` value that exists.
> 
> 
> Maybe I'm not seeing the reasoning correctly. Any thoughts are welcomed indeed.

The process names are something of an implementation detail, and will almost certainly change in the future: once we implement beamforming, an XB-engine will be handling multiple streams, so its name may need to change to reflect that (or possibly we'll have separate X and B katcp device servers within the XB process, in which case the `xb` will become just `x` or `b`). The ICD-defined sensors need to have names that are independent of those implementation details.

As you've noted, we don't have meaningful labels for X engines, so I think `{stream_name}.{i}.{sensor_name}` is a reasonable naming convention.

amishatishpatel:
@bmerry the following are now all the `baseline_correlation_products`-level sensors for a correlator (I manually added empty lines between blocks for ease of differentiation).

```bash
#sensor-list baseline_correlation_products.0.rx-missing-unixtime The timestamp (in UNIX time) when missing data was last detected  timestamp
#sensor-list baseline_correlation_products.0.rx-timestamp The timestamp (in samples) of the last chunk of data received from an F-engine  integer
#sensor-list baseline_correlation_products.0.rx-unixtime The timestamp (in UNIX time) of the last chunk of data received from an F-engine  timestamp
#sensor-list baseline_correlation_products.1.rx-missing-unixtime The timestamp (in UNIX time) when missing data was last detected  timestamp
#sensor-list baseline_correlation_products.1.rx-timestamp The timestamp (in samples) of the last chunk of data received from an F-engine  integer
#sensor-list baseline_correlation_products.1.rx-unixtime The timestamp (in UNIX time) of the last chunk of data received from an F-engine  timestamp
#sensor-list baseline_correlation_products.2.rx-missing-unixtime The timestamp (in UNIX time) when missing data was last detected  timestamp
#sensor-list baseline_correlation_products.2.rx-timestamp The timestamp (in samples) of the last chunk of data received from an F-engine  integer
#sensor-list baseline_correlation_products.2.rx-unixtime The timestamp (in UNIX time) of the last chunk of data received from an F-engine  timestamp
#sensor-list baseline_correlation_products.3.rx-missing-unixtime The timestamp (in UNIX time) when missing data was last detected  timestamp
#sensor-list baseline_correlation_products.3.rx-timestamp The timestamp (in samples) of the last chunk of data received from an F-engine  integer
#sensor-list baseline_correlation_products.3.rx-unixtime The timestamp (in UNIX time) of the last chunk of data received from an F-engine  timestamp

#sensor-list baseline_correlation_products.bandwidth The analogue bandwidth of the digitised band Hz float
#sensor-list baseline_correlation_products.bls-ordering A string showing the output ordering of baseline data produced by the X-engines in this instrument, as a list of correlation pairs given by input label.  string
#sensor-list baseline_correlation_products.channel-data-suspect A bitmask of flags indicating whether each channel should be considered to be garbage.  string
#sensor-list baseline_correlation_products.ddc-mix-freq F-engine DDC mixer frequency, where used. 0 where n/a Hz float
#sensor-list baseline_correlation_products.destination The IP address, range and port to which data for stream baseline_correlation_products is sent  string
#sensor-list baseline_correlation_products.int-time The time, in seconds, for which the X-engines accumulate. s float

#sensor-list baseline_correlation_products.n-accs The number of spectra that are accumulated per X-engine output  integer
#sensor-list baseline_correlation_products.n-bls The number of baselines produced by this correlator instrument.  integer
#sensor-list baseline_correlation_products.n-chans The number of frequency channels in an integration  integer
#sensor-list baseline_correlation_products.n-chans-per-substream Number of channels in each substream for this x-engine stream  integer
#sensor-list baseline_correlation_products.n-xengs The number of X-engines in the instrument  integer

#sensor-list baseline_correlation_products.sync-time The time at which the digitisers were synchronised. Seconds since the Unix Epoch. s float
#sensor-list baseline_correlation_products.xeng-clip-cnt Number of visibilities that saturated  integer
#sensor-list baseline_correlation_products.xeng-out-bits-per-sample X-engine output bits per sample. Per number, not complex pair- Real and imaginary parts are both this wide  integer
#sensor-list baseline_correlation_products.xengs-synchronised For the latest accumulation, was data present from all F-Engines for all X-Engines  boolean
```

bmerry (approved):
Looks good, modulo the suggestion I had on Mattermost about removing the "input." prefix from xbgpu sensors.

PR #652, opened 2023-01-10
Add extra metrics and consul metadata

bmerry:
This is to support a more fully-featured dashboard.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -798,6 +798,9 @@ def _make_fgpu(
         fgpu.pass_telstate = False
         fgpu.data_suspect_sensor = data_suspect_sensor
         fgpu.data_suspect_range = (2 * i, 2 * i + 2)
+        # Identify the antenna, if the input labels are consistent
+        if input_labels[0][:-1] == input_labels[1][:-1]:
+            fgpu.consul_meta["antenna"] = input_labels[0][:-1]
         g.add_node(fgpu)
 
         # Wire it up to the multicast streams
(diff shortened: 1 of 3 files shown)


bmerry:
It's just occurred to me that the dsims might also benefit from the same treatment, although there it is less necessary since for dsims the antenna name gets baked into the task name.

src/katsdpcontroller/generator.py
```diff
@@ -798,6 +798,9 @@ def _make_fgpu(
         fgpu.pass_telstate = False
         fgpu.data_suspect_sensor = data_suspect_sensor
         fgpu.data_suspect_range = (2 * i, 2 * i + 2)
+        # Identify the antenna, if the input labels are consistent
+        if input_labels[0][:-1] == input_labels[1][:-1]:
```
james-smith-za, line 802: Is this just a sanity-check to see whether both pols are coming from the same antenna?

src/katsdpcontroller/generator.py
```diff
@@ -831,10 +834,11 @@ def _make_fgpu(
                 "rx-missing-unixtime",
             ]:
                 fgpu.sensor_renames[f"input{j}-{name}"] = f"{stream.name}-{label}-{name}"
-        # Prepare expected data rate
+        # Prepare expected data rates etc
         fgpu.static_gauges["fgpu_expected_input_heaps_per_second"] = sum(
             src.adc_sample_rate / src.samples_per_heap for src in srcs
         )
+        fgpu.static_gauges["fgpu_expected_engines"] = 1.0
```
james-smith-za, line 841: I'm ... not sure why just 1? Ditto for the xbgpu case down below.

src/katsdpcontroller/generator.py
bmerry: Yes. If an engine is being fed by different antennas (or someone is using it with sources that don't follow the MK input naming convention), avoid producing misleading antenna information.

src/katsdpcontroller/generator.py
bmerry: `fgpu` is a single engine. The number of expected engines per engine is 1. The code that processes static_gauges takes a sum across all the engines.

james-smith-za: Ah. That explains things.

james-smith-za (approved):

PR #654, opened 2023-01-18
raw_antenna_voltage -> baseband_voltage

amishatishpatel:
As per the (draft) CBF-CAM ICD, the stream type name for raw digitiser data has been updated.

Contributes to: NGC-861.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -437,7 +437,7 @@ def _make_meta_writer(
 def _make_dsim(
     g: networkx.MultiDiGraph,
     configuration: Configuration,
-    streams: Iterable[product_config.SimDigRawAntennaVoltageStream],
+    streams: Iterable[product_config.SimDigBasebandVoltageStream],
     sync_time: int,
 ) -> scheduler.LogicalNode:
     """Create the dsim process for a single antenna.
@@ -1009,7 +1009,7 @@ def _make_xbgpu(
 
     input_rate = (
         sum(
-            cast(product_config.DigRawAntennaVoltageStreamBase, dig).adc_sample_rate
+            cast(product_config.DigBasebandVoltageStreamBase, dig).adc_sample_rate
             for dig in acv.src_streams
         )
         * acv.bits_per_sample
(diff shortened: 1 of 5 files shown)


amishatishpatel:
> It may be worth checking with SDP about whether this will break something in their current operation though, because I think if this gets merged, it'll be pushed to site.

Gosh, yes. Didn't even think about those implications!

bmerry:
We should also let CAM know, since this will affect their integration work. @james-smith-za I've forgotten the name of the person in CAM doing the MK+ CBF proxy.

james-smith-za:
> We should also let CAM know, since this will affect their integration work. @james-smith-za I've forgotten the name of the person in CAM doing the MK+ CBF proxy.

It's @lanceWilliams but my main concern was that this may affect MeerKAT operation, will it not?

bmerry (approved):
The code changes look fine to me. I don't think it should affect SDP, because these stream types are not used for SDP subarrays.

A more formal approach would be to create a product-configure schema version 4.0 which uses the new names, and continue to support the old names in schema version 3.x. However I don't think schema 3.1 is really in production anywhere, so unless CAM or SDP object, I'm happy to say that 3.1 is still in development and make a breaking change.

bmerry:
> > We should also let CAM know, since this will affect their integration work. @james-smith-za I've forgotten the name of the person in CAM doing the MK+ CBF proxy.
> 
> It's @lanceWilliams (or possibly @rishadebrahim) but my main concern was that this may affect MeerKAT operation, will it not?

SDP on site hasn't been updated since August; it's not a question of anything we push into katsdpcontroller immediately getting deployed without being tested first.

@lanceWilliams @rishadebrahim are you happy for us to the stream type without bumping the version? It'll just mean that your MK+ CBF proxy will need to do the same rename on the JSON passed to product-configure.

ludwigschwardt:
It looks like the only current users are `katsdpcontroller` and `katgpucbf`. CAM already updated at least their sensors [in 2018](https://skaafrica.atlassian.net/browse/CB-1909).

So I'm happy with this. Supporting both in 3.x and the hypothetical 4.x seems more convoluted.

amishatishpatel:
As per the discussion during this morning's MK+ check-in, the change here is according to the agreed upon CBF-CAM ICD. 
* This change will form part of the continuing development of the 3.1 product-configure schema.

lanceWilliams:
> > > We should also let CAM know, since this will affect their integration work. @james-smith-za I've forgotten the name of the person in CAM doing the MK+ CBF proxy.
> > 
> > 
> > It's @lanceWilliams (or possibly @rishadebrahim) but my main concern was that this may affect MeerKAT operation, will it not?
> 
> SDP on site hasn't been updated since August; it's not a question of anything we push into katsdpcontroller immediately getting deployed without being tested first.
> 
> @lanceWilliams @rishadebrahim are you happy for us to the stream type without bumping the version? It'll just mean that your MK+ CBF proxy will need to do the same rename on the JSON passed to product-configure.

@bmerry I don't have an issue with it, although I admit my understanding of how the stream documentation has been formalized to this point is somewhat vague.

PR #655, opened 2023-01-26
Fix agent-mkconfig for NUMA systems again

bmerry:
Previously ea01ff1 fixed locating PCIe devices for single-NUMA systems, but this caused incorrect output for katsdpcontroller.numa for Epyc systems with NUMA: for NUMA nodes without interesting PCIe devices, it would ascend too far and report all CPU cores of the package as being part of the NUMA node.

### DIFF
--- a/scripts/agent_mkconfig.py
+++ b/scripts/agent_mkconfig.py
@@ -98,12 +98,21 @@ class HWLocParser:
         # cores and devices. We need to ascend the tree to find the container.
         parent_map = {child: parent for parent in self._tree.iter() for child in parent}
         for i in range(len(self._nodes)):
-            if int(self._nodes[i].get("os_index", 0)) != i:
+            node = self._nodes[i]
+            if int(node.get("os_index", 0)) != i:
                 raise RuntimeError("NUMA nodes are not numbered contiguously by the OS")
-            while not self._nodes[i].findall(".//object[@type='PU']") or not self._nodes[i].findall(
-                ".//object[@type='Bridge']"
+            while not node.findall(".//object[@type='PU']"):
+                node = parent_map[node]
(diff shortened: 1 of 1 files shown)


scripts/agent_mkconfig.py
```diff
@@ -98,12 +98,21 @@ def __init__(self):
         # cores and devices. We need to ascend the tree to find the container.
         parent_map = {child: parent for parent in self._tree.iter() for child in parent}
         for i in range(len(self._nodes)):
-            if int(self._nodes[i].get("os_index", 0)) != i:
+            node = self._nodes[i]
+            if int(node.get("os_index", 0)) != i:
                 raise RuntimeError("NUMA nodes are not numbered contiguously by the OS")
-            while not self._nodes[i].findall(".//object[@type='PU']") or not self._nodes[i].findall(
-                ".//object[@type='Bridge']"
+            while not node.findall(".//object[@type='PU']"):
```
amishatishpatel, line 104: Maybe it's just the length of the fully-stitched `node.final(...)` call that makes me uneasy, but
* What do you think about a small helper function to make it look a bit neater?

Likely wholly unnecessary as a change; I just don't like the look of the `findall` arg - so aggressive!

amishatishpatel (approved):
I can't really complain (but I will anyway); just happy that the issue was caught and the change required wasn't extensive.
* Still, it is a very strange accommodation to be doubly-triply sure we're taking care of the 'no interesting bridges' case.

scripts/agent_mkconfig.py
bmerry: Yeah, it was bugging me slightly too. Will make the change.

amishatishpatel (approved):
Excellent.

KimMcAlpine (approved):

PR #656, opened 2023-01-27
Split develop

KimMcAlpine:
Changes to split the current 'develop' mode into finer grained options

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -351,7 +351,7 @@ def _make_telstate(g: networkx.MultiDiGraph, configuration: Configuration) -> sc
     telstate.subsystem = "sdp"
     # redis is nominally single-threaded, but has some helper threads
     # for background tasks so can occasionally exceed 1 CPU.
-    telstate.cpus = 1.2 if not configuration.options.develop else 0.2
+    telstate.cpus = 1.2 if not configuration.options.develop_opts.less_resources else 0.2
     telstate.mem = 2048 + 400 * n_antennas
     telstate.disk = telstate.mem
     telstate.image = "katsdptelstate"
@@ -419,7 +419,9 @@ def _make_meta_writer(
     meta_writer.interfaces = [scheduler.InterfaceRequest("sdp_10g")]
     # Actual required data rate is minimal, but bursty. Use 1 Gb/s,
     # except in development mode where it might not be available.
-    meta_writer.interfaces[0].bandwidth_out = 1e9 if not configuration.options.develop else 10e6
+    meta_writer.interfaces[0].bandwidth_out = (
+        1e9 if not configuration.options.develop_opts.less_resources else 10e6
+    )
     meta_writer.transitions = {
         CaptureBlockState.BURNDOWN: [
             KatcpTransition("write-meta", "{capture_block_id}", True, timeout=120)  # Light dump
@@ -445,7 +447,7 @@ def _make_dsim(
     An antenna has a separate stream per polarisation, so `streams` will
     normally have two elements.
     """
-    ibv = not configuration.options.develop
+    ibv = not configuration.options.develop_opts.disable_ibv
     # dsim assigns digitiser IDs positionally. According to M1000-0001-053,
     # the least significant bit is the polarization ID with 0 = vertical, so
     # sort by reverse of name so that if the streams are, for example,
@@ -490,8 +492,9 @@ def _make_dsim(
     ]
     # Allow dsim task to set a realtime scheduling priority itself
     dsim.taskinfo.container.docker.parameters = [{"key": "ulimit", "value": "rtprio=1"}]
-    if configuration.options.develop:
-        # In develop mode, scale down reservation for low bandwidths to allow
+    if configuration.options.develop_opts.less_resources:
+        # In develop mode with less_resources option=True,
+        # scale down reservation for low bandwidths to allow
         # testing low-bandwidth arrays on a single machine. Use a full core
         # for the maximum sample rate (which is generous - it only needs
         # about 70% for S-band, depending on the CPU).
@@ -533,7 +536,7 @@ def _make_fgpu(
     stream: product_config.GpucbfAntennaChannelisedVoltageStream,
     sync_time: int,
 ) -> scheduler.LogicalNode:
-    ibv = not configuration.options.develop
+    ibv = not configuration.options.develop_opts.disable_ibv
     n_engines = len(stream.src_streams) // 2
     fgpu_group = LogicalGroup(f"fgpu.{stream.name}")
     g.add_node(fgpu_group)
@@ -708,7 +711,7 @@ def _make_fgpu(
         fgpu.fake_katcp_server_cls = FakeFgpuDeviceServer
         fgpu.cpus = 4
         fgpu.mem = 1024  # Actual use is currently around 700 MB
-        if not configuration.options.develop:
+        if not configuration.options.develop_opts.less_resources:
             fgpu.cores = ["src0", "src1", "dst", "python"]
             fgpu.numa_nodes = 1.0  # It's easily starved of bandwidth
             taskset = ["taskset", "-c", "{cores[python]}"]
@@ -771,7 +774,7 @@ def _make_fgpu(
                 "{ports[aioconsole]}",
             ]
         )
-        if not configuration.options.develop:
+        if not configuration.options.develop_opts.less_resources:
             fgpu.command += [
                 "--src-affinity",
                 "{cores[src0]},{cores[src1]}",
@@ -850,7 +853,7 @@ def _make_xbgpu(
     sync_time: int,
     sensors: SensorSet,
 ) -> scheduler.LogicalNode:
-    ibv = not configuration.options.develop
+    ibv = not configuration.options.develop_opts.disable_ibv
     acv = stream.antenna_channelised_voltage
     n_engines = stream.n_substreams
     n_inputs = len(acv.src_streams)
@@ -1050,9 +1053,9 @@ def _make_xbgpu(
         xbgpu.subsystem = "cbf"
         xbgpu.image = "katgpucbf"
         xbgpu.fake_katcp_server_cls = FakeXbgpuDeviceServer
-        xbgpu.cpus = 0.5 * bw_scale if configuration.options.develop else 1.5
+        xbgpu.cpus = 0.5 * bw_scale if configuration.options.develop_opts.less_resources else 1.5
         xbgpu.mem = 512 + _mb(recv_buffer + send_buffer)
-        if not configuration.options.develop:
+        if not configuration.options.develop_opts.less_resources:
             xbgpu.cores = ["src", "dst"]
             xbgpu.numa_nodes = 0.5 * bw_scale  # It's easily starved of bandwidth
             taskset = ["taskset", "-c", "{cores[dst]}"]
@@ -1122,7 +1125,7 @@ def _make_xbgpu(
                 "{ports[aioconsole]}",
             ]
         )
-        if not configuration.options.develop:
+        if not configuration.options.develop_opts.less_resources:
             xbgpu.command += ["--src-affinity", "{cores[src]}", "--dst-affinity", "{cores[dst]}"]
         xbgpu.capabilities.append("SYS_NICE")  # For schedrr
         if ibv:
@@ -1200,7 +1203,7 @@ def _make_cbf_simulator(
     if isinstance(stream, product_config.SimBaselineCorrelationProductsStream):
         while stream.n_vis / n_sim > _N32_32:
             n_sim *= 2
-    ibv = not configuration.options.develop
+    ibv = not configuration.options.develop_opts.disable_ibv
 
     def make_cbf_simulator_config(
         task: ProductPhysicalTask, resolver: "product_controller.Resolver"
@@ -1503,7 +1506,7 @@ def n_cal_nodes(configuration: Configuration, stream: product_config.CalStream) 
     # Use single cal for 4K or less: it doesn't need the performance, and
     # a unified cal report is more convenient (revisit once split cal supports
     # a unified cal report).
-    if configuration.options.develop:
+    if configuration.options.develop_opts.less_resources:
         return 2
     elif stream.vis.n_chans <= 4096:
         return 1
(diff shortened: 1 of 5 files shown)


test/test_product_config.py
```diff
@@ -190,17 +190,28 @@ def test_from_config(self) -> None:
             "service_overrides": {"service1": {"host": "testhost"}},
         }
         options = Options.from_config(config)
-        assert options.develop == config["develop"]
```
ludwigschwardt, line 193: You could probably still check the main "develop" option, right?

test/test_product_config.py
```diff
@@ -190,17 +190,28 @@ def test_from_config(self) -> None:
...
+        assert options.develop_opts.any_gpu is True
+        assert options.develop_opts.disable_ibv is True
+        assert options.develop_opts.less_resources is True
         assert options.wrapper == config["wrapper"]
         assert list(options.service_overrides.keys()) == ["service1"]
         assert options.service_overrides["service1"].host == "testhost"
 
     def test_defaults(self) -> None:
         options = Options.from_config({})
-        assert options.develop is False
```
ludwigschwardt, line 200: Ditto...

ludwigschwardt (approved):
Looks good, just the two minor comments/suggestions.

bmerry:
A high-level comment before I get into the details next week: the way you've designed the scheme means you need to control two different knobs (`develop` and `develop_opts`) and remember how they interact (if I understood correctly, you first turn everything on with `develop`, then turn off things you don't want with `develop_opts`). That's potentially pretty confusing. I also think typical usage is more likely to be turning things on incrementally, rather than off (e.g. "oh it didn't start because we don't have any ibverbs-capable hardware, so turn on the ibverbs disable"). It also means that if we come up with new developer options in future, users have to explicitly request them, rather than getting them automatically when using any other develop options.

What I'd suggest is just keep the one key (`develop`), and allow the value to be either a boolean (which turns on the developer options defined up to this point) OR a dictionary (which can be a set of options to flip on). There is some precedent for this with the `simulate` option in `project_config.json.j2`.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,13 @@
             "additionalProperties": false,
             "properties": {
                 "develop": {"type": "boolean", "default": false},
+{% if version >= "3.2" %}
+		"develop_opts":{
+                    "any_gpu": {"type": "boolean", "default": true},
+                    "disable_ibverbs": {"type": "boolean", "default": true},
```
bmerry, line 617: Double-negatives (`disable_ibverbs: false`) can get confusing. On the other hand, I do like that "false" consistently means the non-developer behaviour. Possibly it could be called `kernel_networking`, but maybe that doesn't make the connection with ibverbs as clearly. So I'm on the fence.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,13 @@
             "additionalProperties": false,
             "properties": {
                 "develop": {"type": "boolean", "default": false},
+{% if version >= "3.2" %}
```
bmerry, line 614: To make a version 3.2, you need to add it to the list of accepted versions at the top of the file.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,13 @@
...
+		"develop_opts":{
```
bmerry, line 615:
```suggestion
		"develop_opts": {
```

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,13 @@
...
+                    "any_gpu": {"type": "boolean", "default": true},
+                    "disable_ibverbs": {"type": "boolean", "default": true},
+                    "less_resources": {"type": "boolean", "default": true}
+                     },
```
bmerry, line 619: Dedent the brace to match the containing scope.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,13 @@
             "additionalProperties": false,
             "properties": {
                 "develop": {"type": "boolean", "default": false},
+{% if version >= "3.2" %}
+		"develop_opts":{
+                    "any_gpu": {"type": "boolean", "default": true},
```
bmerry, line 616: The properties need to go into a "properties" section. I'm a little surprised this is even compiling since "any_gpu" isn't a valid constraint (but then I think json-schema is somewhat permissive).

Also use an `additionalProperties: false` so that users can't add their own developer options that aren't already defined.

src/katsdpcontroller/product_config.py
```diff
@@ -219,18 +219,41 @@ def from_config(cls, config: Mapping[str, Any]) -> "ServiceOverride":
...
         service_overrides: Mapping[str, ServiceOverride] = {},
         interface_mode: bool = False,
     ) -> None:
-        self.develop = develop
+        if not develop:
+            self.develop_opts = DevelopOptions(
+                any_gpu=False, disable_ibv=False, less_resources=False
+            )
+        else:
+            self.develop_opts = DevelopOptions.from_config(develop_opts)
```
bmerry, line 256: What do you think about swapping the two sides of the `if`, so that it's `if develop` rather than `if not develop`.

src/katsdpcontroller/generator.py
```diff
@@ -2206,8 +2213,8 @@ def _make_beamformer_engineering_pol(
         Whether this node is for writing to ramdisk (ignored if `stream` is ``None``)
     idx
         Number of this source within the output (ignored if `stream` is ``None``)
-    develop
-        Whether this is a develop-mode config
+    develop_opts
+        develop_opts of this config
```
bmerry, line 2217: Actually it doesn't seem to be passed at all, so just delete it from the docstring (it's retrieved from `configuration`).

bmerry (changes requested):

KimMcAlpine:
> A high-level comment before I get into the details next week: the way you've designed the scheme means you need to control two different knobs (`develop` and `develop_opts`) and remember how they interact (if I understood correctly, you first turn everything on with `develop`, then turn off things you don't want with `develop_opts`). That's potentially pretty confusing. I also think typical usage is more likely to be turning things on incrementally, rather than off (e.g. "oh it didn't start because we don't have any ibverbs-capable hardware, so turn on the ibverbs disable"). It also means that if we come up with new developer options in future, users have to explicitly request them, rather than getting them automatically when using any other develop options.
> 
> What I'd suggest is just keep the one key (`develop`), and allow the value to be either a boolean (which turns on the developer options defined up to this point) OR a dictionary (which can be a set of options to flip on). There is some precedent for this with the `simulate` option in `project_config.json.j2`.

I'm willing to give this a bash. In my mind I thought it would be less confusing to retain the 'old' behaviour with some additional options, but it is true that there are two knobs to control and it is potentially confusing that the options are meaningless with develop=false.

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: Done

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: removed 'develop_opts' from schema, to simplify design as suggested

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: Done

src/katsdpcontroller/product_config.py
KimMcAlpine: I've modified this so I don't think this comment applies anymore

src/katsdpcontroller/generator.py
KimMcAlpine: Done

test/test_product_config.py
KimMcAlpine: Hmm, I don't think that will work in the new scheme because the `develop` attribute for the `Options` class is no longer a boolean but instead a class with the various options as attributes.

test/test_product_config.py
KimMcAlpine: As above

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: If you are on the fence I'm going to leave it as is then. I agree that the double negatives are confusing but it seems more consistent with the logic of how it was set up originally.

src/katsdpcontroller/product_config.py
```diff
@@ -219,21 +219,46 @@ def from_config(cls, config: Mapping[str, Any]) -> "ServiceOverride":
...
-        develop: bool = False,
+        develop: Union[bool, Mapping[str, bool]] = False,
         wrapper: Optional[str] = None,
         image_tag: Optional[str] = None,
         image_overrides: Mapping[str, str] = {},
         service_overrides: Mapping[str, ServiceOverride] = {},
         interface_mode: bool = False,
     ) -> None:
-        self.develop = develop
+        if type(develop) == bool:
```
bmerry, line 254:
```suggestion
        if isinstance(develop, bool):
```
is more idiomatic

src/katsdpcontroller/product_config.py
```diff
@@ -219,21 +219,46 @@ def from_config(cls, config: Mapping[str, Any]) -> "ServiceOverride":
...
+            self.develop = DevelopOptions.from_bool(develop)
+        else:
+            self.develop = DevelopOptions.from_config(develop)
         self.wrapper = wrapper
         self.image_tag = image_tag
         self.image_overrides = dict(image_overrides)
+        self.image_overrides = dict(image_overrides)
```
bmerry, line 261: Er, what? Copy-paste error?

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,23 @@
             "additionalProperties": false,
             "properties": {
                 "develop": {"type": "boolean", "default": false},
```
bmerry, line 613: You can't have the key present twice. The `if` on the next line should be an if-else. Alternatively, you can put the `if` only around the one branch of the `oneOf` (so that for < 3.2 you get only the boolean branch), which would avoid the need to repeat this line.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,23 @@
...
+{% if version >= "3.2" %}
+                "develop": {
+                    "oneOf": [
+                        {"type": "boolean"},
+                        {
+                            "type": "object",
+                                "properties": {
```
bmerry, line 620:
```suggestion
                            "properties": {
```

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -611,6 +611,23 @@
             "additionalProperties": false,
             "properties": {
                 "develop": {"type": "boolean", "default": false},
+{% if version >= "3.2" %}
+                "develop": {
+                    "oneOf": [
+                        {"type": "boolean"},
```
bmerry, line 617:
```suggestion
                        {"type": "boolean", "default": false},
```

bmerry (changes requested):
The change to a single `develop` key looks good, although I still think it makes more sense to enable develop options one at a time rather than having a giant switch that changes everything after which you can flip smaller switches to undo the effects you didn't want.

There are a few places in the code that still reference version 3.1, that need to be replaced by 3.2. Mostly it's just tests (where it is good to test against the latest version), but this code in product_config.py definitely needs fixing:
```python
    # Upgrade to latest 3.x
    config["version"] = "3.1"
```

KimMcAlpine:
> The change to a single `develop` key looks good, although I still think it makes more sense to enable develop options one at a time rather than having a giant switch that changes everything after which you can flip smaller switches to undo the effects you didn't want.
> 
> There are a few places in the code that still reference version 3.1, that need to be replaced by 3.2. Mostly it's just tests (where it is good to test against the latest version), but this code in product_config.py definitely needs fixing:
> 
> ```python
>     # Upgrade to latest 3.x
>     config["version"] = "3.1"
> ```

I just want to make sure I understand what you mean. Are you suggesting here that the default values of the three options (any_gpu, disable_ibv, and less_resources) should be false ? So if you set 'develop =True' none of these options would be on ?

bmerry:
> > The change to a single `develop` key looks good, although I still think it makes more sense to enable develop options one at a time rather than having a giant switch that changes everything after which you can flip smaller switches to undo the effects you didn't want.
> 
> I just want to make sure I understand what you mean. Are you suggesting here that the default values of the three options (any_gpu, disable_ibv, and less_resources) should be false ? So if you set 'develop =True' none of these options would be on ?

Specifying `"develop:" true` should still turn everything on. I mean that if you provide a dictionary, the individual settings should define to false. So if you decided you wanted to run an array that is larger than you could normally handle and need it to use less resources, you'd do it by saying: `"develop": {"less_resources": true}` rather than by saying `"develop": {"any_gpu": false, "disable_ibverbs": false}` (assuming I've remembered the option names correctly).

src/katsdpcontroller/product_config.py
KimMcAlpine: Done

src/katsdpcontroller/product_config.py
KimMcAlpine: Oops, removed

KimMcAlpine:
> I think I have changed all the 3.1's to 3.2 now

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: I've changed this to put the `if` around the one branch of the `oneOf`

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: Done

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: Done, but I now have to move the `,` to inside the `oneOf` statement, not quite sure how to format this  ?

bmerry: What I think looks the nicest is to swap the order of the two branches of the oneOf, which lets the comma sit in its natural place. But what you've done is fine too.

bmerry (approved):

bmerry (approved):

src/katsdpcontroller/schemas/product_config.json.j2
KimMcAlpine: Thanks, I did consider that as an option. I've changed it.

PR #658, opened 2023-02-02
Prefix ingest command with capambel

richarms:
This change is motivated by a change in the docker daemon and the way linux capabilities are handled. 

CAP_NET_RAW capability is transferred to ingest by `capambel.c`

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1647,7 +1647,8 @@ def _make_ingest(
             ingest.physical_factory = IngestTask
         ingest.fake_katcp_server_cls = FakeIngestDeviceServer
         ingest.image = "katsdpingest_" + normalise_gpu_name(defaults.INGEST_GPU_NAME)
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1647,7 +1647,8 @@ def _make_ingest(
             ingest.physical_factory = IngestTask
         ingest.fake_katcp_server_cls = FakeIngestDeviceServer
         ingest.image = "katsdpingest_" + normalise_gpu_name(defaults.INGEST_GPU_NAME)
-        ingest.command = ["ingest.py"]
+        ingest.command = ["capambel", "-c", "cap_net_raw+p", "-v", "--", "ingest.py"]
```
bmerry, line 1650: The `-v` shouldn't be needed for production - it just spews out debug information.

bmerry (changes requested):

bmerry (approved):

PR #660, opened 2023-02-15
Control autotune matching behavior in ingest tasks

richarms:
This commit sets the KATSDPSIGPROC_TUNE_MATCH environment variable inside ingest containers before launching. This controls the behaviour of the autotuning machinery when an exact match for the GPU device is not found. Setting the variable to "nearest" causes the nearest match to be used, if any exists, while setting it to "exact" forces autotuning for all but exact matches.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -117,6 +117,8 @@ OBJ_DATA_VOL = scheduler.VolumeRequest("obj_data", "/var/kat/data", "RW")
 CONFIG_VOL = scheduler.VolumeRequest("config", "/var/kat/config", "RW")
 #: Number of components in a complex number
 COMPLEX = 2
+#: Autotune fallback behaviour
+KATSDPSIGPROC_TUNE_MATCH = defaults.KATSDPSIGPROC_TUNE_MATCH
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -307,15 +309,19 @@ async def resolve(self, resolver, graph):
...
             gpu_name = normalise_gpu_name(gpu.name)
             image = await resolver.image_resolver("katsdpingest_" + gpu_name)
             if gpu != defaults.INGEST_GPU_NAME:
-                logger.info("Develop mode: using %s for ingest", image.path)
+                logger.info("Develop.any_gpu mode: using %s for ingest", image.path)
         await super().resolve(resolver, graph, image)
+        parameters = self.taskinfo.container.docker.setdefault("parameters", [])
+        parameters.append(
+            {"key": "env", "value": f"KATSDPSIGPROC_TUNE_MATCH={KATSDPSIGPROC_TUNE_MATCH}"}
+        )
```
bmerry, line 324: This is a "low-level" approach to setting an environment variable, where it is put directly onto the Docker command-line. Mesos has a higher-level interface, which I think would look something like this:
```suggestion
        env = self.taskinfo.command.environment.setdefault("variables", [])
        env.append({"name": "KATSDPSIGPROC_TUNE_MATCH", "value": KATSDPSIGPROC_TUNE_MATCH})
```

What's more, I think this can be done in `make_ingest`, alongside setting up the command.

src/katsdpcontroller/generator.py
```diff
@@ -117,6 +117,8 @@ def escape_format(s: str) -> str:
 CONFIG_VOL = scheduler.VolumeRequest("config", "/var/kat/config", "RW")
 #: Number of components in a complex number
 COMPLEX = 2
+#: Autotune fallback behaviour
+KATSDPSIGPROC_TUNE_MATCH = defaults.KATSDPSIGPROC_TUNE_MATCH
```
bmerry, line 121: Is it necessary to copy this from `defaults`, as opposed to just referencing `defaults.KATSDPSIGPROC_TUNE_MATCH` when its used? If it was going to be used many times I can see this would save typing, but since it's only used once it's actually more code rather than less.

bmerry (changes requested):
It looks like this still has a separate Docker image for each GPU. Is the idea just to allow the driver version to vary between tuning and runtime?

src/katsdpcontroller/generator.py
richarms: Accepted change.

src/katsdpcontroller/generator.py
richarms: Thanks for the suggestion. I hadn't found an example of how this had been done previously, but happy to do things this way.

bmerry (approved):

PR #661, opened 2023-03-07
Update sandbox dependencies

bmerry:
Bring all the dependencies in the sandbox up to more recent versions, which should reduce the number of security issues. Thanks to the Mesos update, this now works on an Ubuntu 22.04 host.

I mostly just went for the latest version, but I left the Elastic stack at 7.x (and Kibana at an older 7.x) because Logtrail doesn't support newer versions. Apparently Elastic might have something built-in in newer versions, but that's probably a whole extra kettle of fish to deal with.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
   hooks:
     - id: black
 - repo: https://github.com/PyCQA/isort
-  rev: "5.10.1"
+  rev: "5.12.0"
   hooks:
     - id: isort
 - repo: https://github.com/PyCQA/flake8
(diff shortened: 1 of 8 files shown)


bmerry:
@richarms I think you're probably the only other person who's really used the sandbox. Would you give it a spin and see if it breaks anything for you? If you had a previous sandbox, you'll need to wipe out the minio-data directory, as the newer minio isn't backwards-compatible.

richarms (approved):
I updated docker and installed the new docker-compose-plugin, and the local developer sandbox works on my dev machine with these changes. I haven't tested in the lab.

PR #663, opened 2023-03-27
Add unit tests for agent_mkconfig

bmerry:


bmerry:
Looks like Jenkins is failing because psutil isn't currently in test-requirements. I'll fix that up.

PR #664, opened 2023-03-27
Make list of product-schema versions accessible

bmerry:
It's only accessible internally for now, but could eventually be used to support NGC-906. Additionally, distutils (which is deprecated) is replaced by packaging.version. Because packaging.version.Version doesn't automatically implement comparisons to stringified versions the way distutils.version.StrictVersion did, I had to write a ComparableVersion wrapper.

Instead of individual schema files defining a validate_version schema, they simply list the versions they support, and the validation schema is generated. This also allows each schema object to report its versions e.g. PRODUCT_CONFIG.versions. As noted, this facility isn't wired up to anything yet.

### DIFF
--- a/src/katsdpcontroller/schemas/__init__.py
+++ b/src/katsdpcontroller/schemas/__init__.py
@@ -2,13 +2,18 @@
 
 import codecs
 import json
-from distutils.version import StrictVersion
 
 import jinja2
 import jsonschema
 import pkg_resources
+from packaging.version import Version
 
 _env = jinja2.Environment(loader=jinja2.PackageLoader(__name__, "."))
+
+
+def _normalise_version(version):
+    """Coerce strings to ComparableVersion."""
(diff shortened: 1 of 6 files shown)


src/katsdpcontroller/schemas/__init__.py
```diff
@@ -2,48 +2,81 @@
...
+
+
 def _make_validator(schema):
     """Check a schema document and create a validator from it"""
     validator_cls = jsonschema.validators.validator_for(schema)
     validator_cls.check_schema(schema)
     return validator_cls(schema, format_checker=jsonschema.FormatChecker())
 
 
+class ComparableVersion(Version):
```
ludwigschwardt, line 26: How about using `functools.total_ordering`?

bmerry: I thought of that, but according to the docs it won't override comparators that are already implemented in a base class.

PR #667, opened 2023-03-28
Add hyphen to allowed characters in stream-name

james-smith-za:
CAM would like to be able to call CBF streams things like `antenna-channelised-voltage`, rather than
`antenna_channelised_voltage`. I think this commit should enable it.

(Possibly) closes NGC-914.

### DIFF
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -2074,7 +2074,7 @@ def _upgrade(config):
         config["version"] = "3.0"
 
     # Upgrade to latest 3.x
-    config["version"] = schemas.PRODUCT_CONFIG.versions[-1]
+    config["version"] = str(schemas.PRODUCT_CONFIG.versions[-1])
 
     _validate(config)  # Should never fail if the input was valid
     return config
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -2074,7 +2074,7 @@ def _upgrade(config):
         config["version"] = "3.0"
 
     # Upgrade to latest 3.x
-    config["version"] = "3.3"
+    config["version"] = schemas.PRODUCT_CONFIG.versions[-1]
 
     _validate(config)  # Should never fail if the input was valid
     return config
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -2074,7 +2074,7 @@ def _upgrade(config):
         config["version"] = "3.0"
 
     # Upgrade to latest 3.x
-    config["version"] = str(schemas.PRODUCT_CONFIG.versions[-1])
+    config["version"] = str(max(schemas.PRODUCT_CONFIG.versions))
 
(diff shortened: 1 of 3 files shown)


james-smith-za:
Ok, Jenkins fails with an ominous "Failed to process config: Invalid \\escape: line 8 column 37 (char 174)".
So clearly it doesn't like my escape of the hyphen for some reason.

ludwigschwardt:
Maybe add a `r` in front of the string, since the backslash is meant for `re` instead?

james-smith-za:
> Maybe add a `r` in front of the string, since the backslash is meant for `re` instead?
> 
> In other words: `r"^[A-Za-z0-9_\-]+$"`.
> 
> Oops sorry, scratch that - it's not Python but JSON :-)

Unfortunately I didn't see your edit before I pushed a change. I give up for the afternoon though.

src/katsdpcontroller/product_config.py
```diff
@@ -2074,7 +2074,7 @@ def _upgrade(config):
         config["version"] = "3.0"
 
     # Upgrade to latest 3.x
-    config["version"] = "3.3"
+    config["version"] = str(schemas.PRODUCT_CONFIG.versions[-1])
```
bmerry, line 2077: I remember this is what I suggested, but now that I see it I wonder if
```suggestion
    config["version"] = str(max(schemas.PRODUCT_CONFIG.versions))
```
captures the intent better.

bmerry (approved):
Looks good. One bit of bikeshedding which you can feel free to apply or not.

bmerry:
Another bit of optional bikeshedding: if you move the `-` to the end of the character class instead of the beginning, it's clearer that both `_` and `-` are available.

src/katsdpcontroller/product_config.py
james-smith-za: Does this interpret the strings properly as float numbers? I'm a little hazy on how the parsing works from j2 template to json to python.

bmerry: schemas.PRODUCT_CONFIG.versions are instances of `packaging.version.Version`, which sort the way versions should (see `katsdpcontroller.schemas._normalise_version`).

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -1,15 +1,16 @@
 {% macro versions() %}
-["2.4", "2.5", "2.6", "3.0", "3.1", "3.2", "3.3"]
+["2.4", "2.5", "2.6", "3.0", "3.1", "3.2", "3.3", "3.4"]
 {% endmacro %}
 
 {% macro validate(version) %}
 {% set input_additional = "false" if version >= "3.0" else "true" %}
+{% set stream_name_pattern = "^[-A-Za-z0-9_]+$" if version >= "3.4" else "^[A-Za-z0-9_]+$" %}
```
ludwigschwardt, line 7: Thanks, now I've (re-)learnt something about re. 😄

ludwigschwardt (approved):

PR #669, opened 2023-03-29
Pass narrowband arguments to fgpu

bmerry:
This won't actually work yet since fgpu doesn't yet support narrowband, but once fgpu does support narrowband the support for it should be in place and it won't accidentally start a wideband instead.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -532,9 +532,6 @@ def _make_fgpu(
     stream: product_config.GpucbfAntennaChannelisedVoltageStream,
     sync_time: int,
 ) -> scheduler.LogicalNode:
-    if stream.narrowband is not None:
-        raise NotImplementedError("narrowband support isn't implemented yet")
-
     ibv = not configuration.options.develop.disable_ibv
     n_engines = len(stream.src_streams) // 2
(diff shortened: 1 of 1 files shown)


james-smith-za (commented):
Would you not need multiple such `output_config` dictionaries?

I may be wrong but the picture in my mind that I have is that you'd pass something like 
```
--wideband=channels=12,taps=7,dst=239.256.257.258-14:9876
--narrowband=<etc>
--narrowband=<another etc potentially>
```

?

bmerry:
> Would you not need multiple such `output_config` dictionaries?

The end goal would be to have generator.py identify antenna-channelised-voltage streams that have the same set of sources (and are generally compatible) and merge them into one set of F-engines with multiple `--wideband`/`--narrowband` args, but that's further down the road (specifically, https://skaafrica.atlassian.net/browse/NGC-916).

> 
> I may be wrong but the picture in my mind that I have is that you'd pass something like
> 
> ```
> --wideband=channels=12,taps=7,dst=239.256.257.258-14:9876
> --narrowband=<etc>
> --narrowband=<another etc potentially>
> ```
> 
> ?

james-smith-za (approved):

PR #671, opened 2023-03-30
Pass stream name to ?gain, ?gain-all, ?delays

bmerry:
This is to match changes to fgpu to take the stream name as an argument.

See NGC-569.

### DIFF
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -1284,7 +1284,15 @@ class SubarrayProduct:
         msgs = []
         for i in range(n_inputs // 2):
             nodes.append(self._nodes[f"f.{stream.name}.{i}"])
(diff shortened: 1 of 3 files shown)


src/katsdpcontroller/product_controller.py
amishatishpatel, line 1: Unfortunately I'm going to be a bit of a stickler on this `stream` vs `stream_name` discussion. I'm sure it's always been `stream_name` at the Product level and I just haven't noticed.

Still, will keep an eye out for your response on ska-sa/katgpucbf#544

amishatishpatel (commented):
All looks good, just want/waiting to clarify `stream` vs `stream_name` before explicitly approving.

src/katsdpcontroller/product_controller.py
bmerry: I've renamed them.

amishatishpatel (approved):

PR #672, opened 2023-03-30
SPR1-2376: Don't dump RDB files if not archiving

ludwigschwardt:
While the meta_writer's `write_meta` KATCP request is quite flexible in terms of stream input, it is only ever called with the default setting, which looks for archived `sdp.vis` streams. The resulting stream names are appended to the CBID to form corresponding RDB filenames.

When testing in the lab, we frequently disable archiving streams to save disk space. The side effect is that meta_writer crashes with "No stream specified, and cannot determine available streams from telstate." This happens right at the end of the observation so it's not that critical, but it leaves a bad taste in the mouth.

Avoid this crash by disabling RDB dumps if there are no archived streams. This is necessary but not sufficient, but should at least get rid of the typical lab crashes.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -2502,9 +2502,13 @@ def build_logical_graph(
                 need_telstate = True
     if need_telstate:
         telstate: Optional[scheduler.LogicalNode] = _make_telstate(g, configuration)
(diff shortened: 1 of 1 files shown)


ludwigschwardt:
I guess people would like it for debugging purposes, but there are ways to keep the subarray running after a kattelmod script (the `--dont-stop` option), which allows you to interrogate Redis directly.

KimMcAlpine (approved):

PR #673, opened 2023-03-31
Add product-configure-versions request

bmerry:
This reports the product configure schema versions that are supported, assuming that nothing in the config dict overrides the choice of product controller.

Closes NGC-906 (although the ICD will still need to be updated).


james-smith-za:
Is there a strong reason to have made this a request and not a sensor?

bmerry:
> Is there a strong reason to have made this a request and not a sensor?

The master controller has no way of knowing when the information changes - it has to be explicitly queried.

james-smith-za (approved):

PR #674, opened 2023-04-03
Fix 'less_resources' without 'disable_ibv'

bmerry:
It was causing a KeyError in fgpu/xbgpu because it tried to use core numbers to set ibv completion vectors, but no cores were reserved.

Closes NGC-917.


ludwigschwardt (approved):
If you don't set the completion vectors, what impact does it have on ibverbs support?

bmerry:
> If you don't set the completion vectors, what impact does it have on ibverbs support?

They'll all default to 0, which will put an unbalanced load on one CPU core. But this mode is only supposed to be used for low-bandwidth scenarios anyway.

james-smith-za (approved):

PR #675, opened 2023-04-05
Move fgpu adc-bits sensor to f-engine stream grouping

james-smith-za:
This action taken in order to be more consistent with adc-sample-rate.

Closes NGC-920


james-smith-za:
I thought there would be more to it than just this.

bmerry (approved):
Nice and easy :-)

PR #679, opened 2023-04-20
Merge fgpu streams

bmerry:
When multiple gpucbf.antenna_channelised_voltage streams have the same
input configuration (just different outputs), run them in the same
engines.

This partially addresses NGC-916. There are a few outstanding issues:
- Resource usage may need adjustment (it will use more memory, for
  example).
- A few sensors need to be replicated across streams, which will require
  some rework of the sensor renaming apparatus.


bmerry:
This one has a lot of changes that are just due to indentation; it may be worth reviewing with whitespace changes ignored.

james-smith-za (approved):

james-smith-za (approved):
Ah, ok. Now I see the use-case.

PR #680, opened 2023-04-26
SPR1-2376: Let dsim set sub_band telstate entry

ludwigschwardt:
In the absence of CAM we need to set certain telstate keys as part of product configuration. One is `sub_band`, needed by cal to start up. The CBF simulator stream ("sim.cbf") already initialises `sub_band`, so do the same for the lab test setup of the GPU correlator ("gpucbf").

Dsim is a good place to put this because
  - it knows the band,
  - it already pre-populates `mxxx_observer` in the absence of CAM, and
  - it is typically not needed if CAM is around.

This change allows cal to start up in the lab.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -457,6 +457,7 @@ def _make_dsim(
     telstate_data = {"observer": streams[0].antenna.description}
     for key, value in telstate_data.items():
         init_telstate[(streams[0].antenna.name, key)] = value
+    init_telstate[("sub", "band")] = streams[0].band
 
     # Generate a unique name. The caller groups streams by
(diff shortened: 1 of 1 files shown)


bmerry (commented):
This PR certainly has the benefit of being minimal. I'm just wondering about a hypothetical SDP-CBF-DIG integration (no CAM) using a real digitiser (or SKARAB dsim), which will land us with the same problem again. It may be better to add the code to `_build_logical_graph` directly, and to consider both dig.baseband_voltage and sim.dig.baseband_voltage streams (and perhaps also to skip it if there is a cam.http stream, to avoid having multiple sources of truth for telstate).

bmerry (approved):
Approving since I have no fundamental objection to this approach, despite the bikeshedding suggestions.

PR #681, opened 2023-05-02
Bump flask, dash, werkzeug to newer versions

bmerry:
This started out as bumping flask and werkzeug to keep dependabot happy (security advisories), but required updating a bunch of other packages in the stack due to incompatibilities. The good news is that it got rid of the dependency on ipython. The bad news is that it required overriding some package versions from dockerbase because I don't want to deal with the fallout of updating dockerbase and breaking half the downstream packages.

Relates to SPR1-2783.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,13 @@
--c https://raw.githubusercontent.com/ska-sa/katsdpdockerbase/master/docker-base-build/base-requirements.txt
+-d https://raw.githubusercontent.com/ska-sa/katsdpdockerbase/master/docker-base-build/base-requirements.txt
 
 addict==2.1.1
 aiohttp
-aiohttp-jinja2==1.4.2
+aiohttp-jinja2==1.5.1
 aiokatcp
 aiozk==0.14.0
 async-timeout
-click==7.0                 # via Flask
+blinker==1.6.2             # via flask
+click==8.1.3               # via flask
 dash==1.20.0
 dash_core_components==1.16.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,50 +1,49 @@
...
+importlib-metadata==6.6.0  # via flask
+itsdangerous==2.1.2        # via flask
+jinja2==3.1.2              # override old version in katsdpdockerbase
 jsonschema
-jupyter-core==4.11.2       # via nbformat
 katportalclient
 kazoo==2.8.0
+markupsafe==2.1.2          # via jinja2, werkzeug - override old version in katsdpdockerbase
 monotonic==1.5             # via prometheus_async
-nbformat==4.4.0            # via plotly
```
ludwigschwardt, line 30: Are we not plotting notebooks or does `plotly` not require it anymore?

requirements.txt
```diff
@@ -1,50 +1,49 @@
...
 prometheus_async==18.1.0
 prometheus_client==0.3.1
 pydot==1.4.2
 pymesos==0.3.6
-retrying==1.3.3            # via plotly
 rfc3987==1.3.7
-traitlets==4.3.2           # via nbformat
+tenacity==8.2.2            # via plotly
 websocket-client==0.45.0   # via docker
-werkzeug==0.15.5           # via flask
```
ludwigschwardt, line 44: Quite a jump :-)

ludwigschwardt (approved):

PR #682, opened 2023-05-02
Improve sorting on the dashboard

bmerry:
The default order of tasks on the dashboard annoyed me (it interleaved fgpu and dsim tasks, and sorted 10 before 2), and the sort buttons in the UI were broken. This should improve both.

Before:
Before

After:
After


james-smith-za (approved):
I'm not terribly well educated on how networkx stuff works.

But I'm quite happy with the outcome, I like it.

PR #687, opened 2023-05-19
Remove use of deprecated pkg_resources

bmerry:
Use importlib_resources instead (can eventually be replaced by stdlib importlib.resources once Python 3.8 support is no longer needed).

See NGC-977.


james-smith-za (approved):
LGTM

PR #690, opened 2023-06-01
Slight adjustment to fgpu GPU memory usage

bmerry:
There have been various recent changes to fgpu, and this brings the
allocation up to date with actual usage.


james-smith-za (commented):
Is 1/8 using more memory than 1/16 mode?

bmerry:
> Is 1/8 using more memory than 1/16 mode?

Yes, because you have more intermediate data. I'm not sure why the comment originally referred to 1/16 mode.

james-smith-za (approved):
I was under the impression that 1/16 needed a bit more because you needed to hold more input data in order to make an entire output spectrum, due to the decimation.

bmerry:
> I was under the impression that 1/16 needed a bit more because you needed to hold more input data in order to make an entire output spectrum, due to the decimation.

That's true, and I guess perhaps previously that caused 1/16 to use more. But when I tested the latest version it seemed that the balance is that 1/8 is the bigger memory user.

PR #691, opened 2023-06-01
agent_mkconfig: use /etc/mesos-agent if it exists

bmerry:
This is the newer location, superseding /etc/mesos-slave. If /etc/mesos-agent doesn't exist, fall back to the original.


bmerry:
@madhobbez it seems like the copies of this script in katsdpcontroller and in katsdpinfrastructure drifted out of sync. This is an attempt to bring them back into sync by porting the changes you made in katsdpinfrastructure over to katsdpcontroller, after which I'll just copy the script to katsdpinfrastructure.

madhobbez (approved):
LGTM

PR #692, opened 2023-06-05
Add --corrprod argument for xbgpu

amishatishpatel:
Doesn't need two reviewers. Whoever gets to it first, please and thank you.

Contributes to: NGC-905.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1172,7 +1172,7 @@ def _make_xbgpu(
         )
         output_config = {
             "name": escape_format(stream.name),
-            "heap_accumulation_threshold": str(round(stream.int_time / heap_time)),
+            "heap_accumulation_threshold": round(stream.int_time / heap_time),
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1172,6 +1170,15 @@ def get_baseline_index(a1, a2):
                 "{ports[aioconsole]}",
             ]
         )
+        output_config = {
+            "name": escape_format(stream.name),
+            "heap_accumulation_threshold": str(round(stream.int_time / heap_time)),
```
bmerry, line 1175:
```suggestion
            "heap_accumulation_threshold": round(stream.int_time / heap_time),
```
The `str` isn't needed because the `f"{key}=value"` lower down will stringify it.

bmerry (changes requested):
Looks good. Just one bit of bike-shedding.

bmerry (approved):

PR #696, opened 2023-06-14
Improve diagnosis when there are not enough resources

bmerry:
Use a more powerful approach to isolating a single reason that a set of tasks cannot be launched. See NGC-997 and the docstring of the diagnose_insufficient module for more details.

The functionality was getting a bit too big and complex to be part of the Scheduler class, so it's split into a new module (and the tests are also moved to a new matching test module).

### DIFF
--- a/src/katsdpcontroller/diagnose_insufficient.py
+++ b/src/katsdpcontroller/diagnose_insufficient.py
@@ -469,6 +469,10 @@ def diagnose_insufficient(
         # Non-tasks aren't relevant, so filter them out.
         tasks = [node for node in nodes if isinstance(node, PhysicalTask)]
 
+        # First use a weak filter. This avoids an error saying that a task
+        # require more than zero of a resource of which zero was available,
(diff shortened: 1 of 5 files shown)


james-smith-za (approved):
I'm not really able to make any useful comments on this.

The explanation seems reasonable to me, and it doesn't look as though it's breaking anything based on the tests.

I'd like to see the new scheduler in action, and deliberately try to fail some things to see what difference it makes in error messages.

bmerry:
> I'd like to see the new scheduler in action, and deliberately try to fail some things to see what difference it makes in error messages.

The case where I've observed a difference so far is that previously when trying to start a correlator that needed too much memory for the cbfgpu* nodes, it would give a generic error, because it considered GPU memory on *all* nodes, including `gareth`. The new version considers only the nodes that are able to run fgpu/xbgpu (which excludes gareth, because we've deliberately not told Mesos about its high-speed NIC so that it won't use it) and so it was able to report that those nodes didn't have enough GPU memory.

PR #701, opened 2023-07-21
Fix disable_ibverbs develop option

bmerry:
Internally it was being looked up as `config.get("disable_ibv", False`, which was always false because the actual name is "disable_ibverbs".

Renamed to disable_ibverbs throughout for consistency.


bmerry:
@KimMcAlpine would you mind giving this a sanity-check since you wrote this code?

KimMcAlpine (approved):
Thanks for catching this

PR #703, opened 2023-07-31
Add sensors to indicate network interfaces used

bmerry:
Relates to NGC-1017.


james-smith-za (approved):
I guess my only follow-up question would be, have you tested this in the high-bandwidth case where you're using more than one network interface per engine?

bmerry:
> I guess my only follow-up question would be, have you tested this in the high-bandwidth case where you're using more than one network interface per engine?

No. katsdpcontroller currently doesn't support more than one interface per network (it uses a dictionary to map network name to the interface information).

PR #704, opened 2023-07-31
Interact asynchronously with the Mesos driver

bmerry:
Calls to self._driver.foo() make an HTTP request synchronously, which can block the event loop. This was particularly problematic with the reconciliation, which was blocking the event loop during ?delay calls.

Fix it by creating an AsyncDriver wrapper that turns synchronous calls into fire-and-forget submissions to a ThreadPoolExecutor. This does have the unwanted side effect that failures from Mesos will be dropped rather than fed back to the calling code, but in most cases the failures are delivered asynchronously via a status notification anyway.

Closes NGC-1013.

### DIFF
--- a/src/katsdpcontroller/scheduler.py
+++ b/src/katsdpcontroller/scheduler.py
@@ -223,6 +223,7 @@ import base64
 import contextlib
 import copy
 import decimal
+import functools
 import io
 import ipaddress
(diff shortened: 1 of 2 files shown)


james-smith-za (approved):
Clever. I like it.

PR #705, opened 2023-08-24
xbgpu sensor renames and capture-{start, stop} updates

amishatishpatel:
Update xbgpu sensors and capture-{start, stop} request according to https://github.com/ska-sa/katgpucbf/pull/600
* The XPipeline-specific sensors have been abstracted here to be part of the **stream**, naming as follows
  * `<stream_name>.<xeng_id>.<sensor_name>`, e.g. `baseline-correlation-products.1.rx.synchronised`
* The capture-{start, stop} commands are still invoked the same way, `?capture-start <stream_name>`
   * But now the function handling the request makes the updated call to the XBEngine using the `stream_name` arg again
   * As previously agreed, at least in the case of the `baseline-correlation-products` stream, the Pipeline name is the same as the top-level stream name, so we can reuse the variable (i.e. not go digging further).

I've purposefully left out any logic required for eventual beamformer/tied-array-channelised-voltage streams as I'm not yet sure if there will be changes required to katgpucbf.
* Still, the (re)use of the `stream_name` is clear enough now that it will likely be unchanged in the beamformer-specific case(s).

Resolves: NGC-1049 and NGC-1057.

### DIFF
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -34,6 +34,11 @@ def _format_complex(value: numbers.Complex) -> str:
     return f"{value.real}{value.imag:+}j"
 
 
+def _parse_key_val(value: str) -> Dict[str, str]:
+    """Turn ``"foo=a,bar=b"`` into ``{"foo": "a", "bar": b"}``."""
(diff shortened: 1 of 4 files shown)


bmerry (approved):
All looks pretty sensible. And nice to see that the chan-range and xeng-clip-cnt sensors are now getting aligned to the ICD.

PR #709, opened 2023-12-06
Add parsing for product-configure 3.5

amishatishpatel:
It parses a GpucbfTiedArrayChannelisedVoltageStream config object, but doesn't
actually do anything with it. Trying to instantiate a beamformer with it will result in
a `NotImplementedError` being raised.

See suggested edits in the doc below for the schema update.
- https://docs.google.com/document/d/15ZIyuBf4Vk4ESIEdKsQxdSBoLEwh85eyKeDJouLRAZY/edit?usp=sharing

Closes: NGC-932 and NGC-933.

### DIFF
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -1257,7 +1257,7 @@ class GpucbfTiedArrayChannelisedVoltageStream(TiedArrayChannelisedVoltageStreamB
     if TYPE_CHECKING:  # pragma: nocover
         # Refine the return type for mypy
         @property
-        def antenna_channelised_voltage(self) -> AntennaChannelisedVoltageStreamBase:
+        def antenna_channelised_voltage(self) -> GpucbfAntennaChannelisedVoltageStream:
             ...
 
     @classmethod
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -1226,6 +1226,55 @@ class TiedArrayChannelisedVoltageStream(CbfStream, TiedArrayChannelisedVoltageSt
             instrument_dev_name=config["instrument_dev_name"],
         )
 
+class GpucbfTiedArrayChannelisedVoltageStream(TiedArrayChannelisedVoltageStreamBase):
+    """Real tied-array-channelised-voltage stream (GPU beamformer)."""
+
+    stream_type: ClassVar[str] = "gpucbf.tied_array_channelised_voltage"
+    _valid_src_types: ClassVar[_ValidTypes] = {"gpucbf.antenna_channelised_voltage"}
+
+    def __init__(
+        self,
+        name: str,
+        src_streams: Sequence[Stream],
+        *,
+        src_pol: int,
+        command_line_extra: Iterable[str] = (),
+    ) -> None:
+        acv = src_streams[0]
+        assert isinstance(acv, GpucbfAntennaChannelisedVoltageStream)
+        super().__init__(
+            name,
+            src_streams,
+            n_endpoints=acv.n_substreams,
+            n_chans_per_substream=acv.n_chans_per_substream,
+            spectra_per_heap=acv.n_spectra_per_heap,
+            bits_per_sample=8,
+        )
+        self.src_pol = src_pol
+        self.command_line_extra = list(command_line_extra)
+
+    if TYPE_CHECKING:   # pragma: nocover
+        # Refine the return type for mypy
+        @property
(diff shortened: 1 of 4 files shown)


src/katsdpcontroller/product_config.py
```diff
@@ -1227,6 +1227,56 @@ def from_config(
...
+            spectra_per_heap=acv.n_spectra_per_heap,
+            bits_per_sample=8,
+        )
+        self.src_pol = src_pol
+        self.command_line_extra = list(command_line_extra)
+
+    if TYPE_CHECKING:  # pragma: nocover
+        # Refine the return type for mypy
+        @property
+        def antenna_channelised_voltage(self) -> AntennaChannelisedVoltageStreamBase:
```
bmerry, line 1260: I think this should be
```suggestion
        def antenna_channelised_voltage(self) -> GpucbfAntennaChannelisedVoltageStream:
```
Otherwise it's not adding any value

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -603,6 +606,22 @@
...
                                 "required": ["src_streams", "int_time"]
                             }
+                        },
+                        {
+                            "if": {"properties": {"type": {"const": "gpucbf.tied_array_channelised_voltage"}}},
+                            "then": {
+                                "properties": {
+                                    "type": {},
+                                    "src_streams": {"$ref": "#/definitions/singleton"},
+                                    "src_pol": {"$ref": "#/definitions/positive_integer"},
```
bmerry, line 616: positive_integer is definitely wrong, since it disallows 0. It should actually be stricter and only allow 0 or 1. I'm rusty on the syntax, but I think something like
```suggestion
                                    "src_pol": {"type": { "enum": [0, 1] }},
```

bmerry (changes requested):
I'm basically happy with this - just a couple of small fixes.

There isn't currently any test coverage for the JSON schema (as evidenced by the use of positive_integer to constrain src_pol), but that'll come naturally when testing for NGC-1116.

james-smith-za (commented):
Looks fine to me, other than the comments that Bruce has made.

src/katsdpcontroller/schemas/product_config.json.j2
amishatishpatel: I'd originally uhm'd and ah'd between `positive_integer` and `positive_number`, but I figured the latter was more for float-types. I admittedly couldn't find anything online Re: syntax, but appreciate the heads-up - will do more digging to confirm your suggestion works.

amishatishpatel: Didn't have to look too far to confirm syntax, there are existing uses in this file. One of particular interest/inspiration is around line 164, under `cbf.antenna_channelised_voltage`
* `"n_pols": {"type": "integer", "enum": [1, 2]},`

Ignoring the actual values, I see the `"integer"` field/identifier is optional; other usages simply follow
* `"field_name": { "type": {"enum": ["value1", "value2", ... ]} }`

I gather the inclusion of `integer` in the first excerpt/example meant the options had to start from 1 rather than 0. I think I'm going to stick with your original suggestion; please and thank you once again.

bmerry: Actually, the second usage is invalid syntax - perhaps you can correct them. The value of `type` should always be a string or a list of strings, each of which is a primitive type (see the spec [here](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01#rfc.section.6.1)).

For `n_pols`, the `type` is mostly redundant, because if a value equals 1 or 2 it must be an integer. However, I think the Python json_schema package may use it to disallow `1.0` (JSON itself doesn't distinguish between integers and integral floats), so it probably doesn't hurt to have it. So I think the correct thing to use here would be
```json
"src_pol": {"type": "integer", "enum": [0, 1]}
```

bmerry: > Actually, the second usage is invalid syntax - perhaps you can correct them.

Ignore me - I see what's going on. It's valid syntax in its context: in the context, the `"type"` isn't a constraint (saying that a field must have a particular type), it's the *name* of a property in the document to be validated i.e. a stream must have a key called `"type"` and it's value must adhere to the `"enum"` restriction defined within.

amishatishpatel: I'm going to try your suggestion (below) and poke sdp-jenkins to build it.
```JSON
"src_pol": {"type": "integer", "enum": [0, 1]}
```

PR #714, opened 2024-02-09
Add gpucbf B-engine sensors

amishatishpatel:
Also add 
* a SumSensor for beng-clip-cnt sensors at the Engine level - like xeng-clip-cnt, and
* a missing scale-factor-timestamp sensor.

Re: Proof this actually works
* The first screenshot/excerpt below is of a `?sensor-list` on a node running one `--corrprod` and two `--beam`s
* The second excerpt is from a `?sensor-list` on the product itself, to show the stream-level sensors

### One XB-engine's sensors
image

```bash
#sensor-list baseline-correlation-products.chan-range The range of channels processed by this X-engine, inclusive  string
< #sensor-list baseline-correlation-products.rx.synchronised For the latest accumulation, was data present from all F-Engines.  boolean
< #sensor-list baseline-correlation-products.xeng-clip-cnt Number of visibilities that saturated  integer
< #sensor-list device-status Overall engine health  discrete ok degraded fail
< #sensor-list rx.device-status XB-engine is receiving a good, clean F-engine stream  discrete ok degraded fail
< #sensor-list rx.missing-unixtime The timestamp (in UNIX time) when missing data was last detected  timestamp
< #sensor-list rx.timestamp The timestamp (in samples) of the last chunk of data received from an F-engine  integer
< #sensor-list rx.unixtime The timestamp (in UNIX time) of the last chunk of data received from an F-engine  timestamp
< #sensor-list tied-array-channelised-voltage0x.beng-clip-cnt Number of complex samples that saturated.  integer
< #sensor-list tied-array-channelised-voltage0x.chan-range The range of channels processed by this B-engine, inclusive  string
< #sensor-list tied-array-channelised-voltage0x.delay The delay settings of the inputs for this beam: (loadmcnt <ADC sample count when model was loaded>, delay <in seconds>, phase <radians>, ...)  string
< #sensor-list tied-array-channelised-voltage0x.quantiser-gain Non-complex post-summation quantiser gain applied to this beam  float
< #sensor-list tied-array-channelised-voltage0x.weight The summing weights applied to all the inputs of this beam  string
< #sensor-list tied-array-channelised-voltage0y.beng-clip-cnt Number of complex samples that saturated.  integer
< #sensor-list tied-array-channelised-voltage0y.chan-range The range of channels processed by this B-engine, inclusive  string
< #sensor-list tied-array-channelised-voltage0y.delay The delay settings of the inputs for this beam: (loadmcnt <ADC sample count when model was loaded>, delay <in seconds>, phase <radians>, ...)  string
< #sensor-list tied-array-channelised-voltage0y.quantiser-gain Non-complex post-summation quantiser gain applied to this beam  float
< #sensor-list tied-array-channelised-voltage0y.weight The summing weights applied to all the inputs of this beam  string
```

### Stream-level sensors
image

```bash
< #sensor-list tied-array-channelised-voltage0x.bandwidth The analogue bandwidth of the digitised band Hz float
< #sensor-list tied-array-channelised-voltage0x.beng-clip-cnt Number of visibilities that saturated  integer
< #sensor-list tied-array-channelised-voltage0x.beng-out-bits-per-sample X-engine output bits per sample. Per number, not complex pair- Real and imaginary parts are both this wide  integer< #sensor-list tied-array-channelised-voltage0x.channel-data-suspect A bitmask of flags indicating whether each channel should be considered to be garbage.  string
< #sensor-list tied-array-channelised-voltage0x.destination The IP address, range and port to which data for stream tied-array-channelised-voltage0x is sent  string
< #sensor-list tied-array-channelised-voltage0x.n-chans The number of frequency channels in this stream's overall output  integer
< #sensor-list tied-array-channelised-voltage0x.n-chans-per-substream Number of channels in each substream for this data stream  integer
< #sensor-list tied-array-channelised-voltage0x.scale-factor-timestamp Factor by which to divide instrument timestamps to convert to unix seconds Hz float
< #sensor-list tied-array-channelised-voltage0x.source-indices The global input indices of the sources summed in this beam  string
< #sensor-list tied-array-channelised-voltage0x.spectra-per-heap Number of spectra per heap of beamformer output  integer
< #sensor-list tied-array-channelised-voltage0x.sync-time The time at which the digitisers were synchronised. Seconds since the Unix Epoch. s float
```

Resolves: NGC-447.

### DIFF
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2013-2023, National Research Foundation (SARAO)
+# Copyright (c) 2013-2024, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
@@ -267,10 +267,11 @@ class FakeFgpuDeviceServer(FakeDeviceServer):
 class FakeXbgpuDeviceServer(FakeDeviceServer):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
+        antennas = self.get_command_argument(int, "--array-size")
         channel_offset = self.get_command_argument(int, "--channel-offset-value")
         channels_per_substream = self.get_command_argument(int, "--channels-per-substream")
-        beam_outpus = self.get_command_arguments(_parse_key_val, "--beam")
-        beam_names = [beam["name"] for beam in beam_outpus]
+        beam_outputs = self.get_command_arguments(_parse_key_val, "--beam")
+        beam_names = [beam["name"] for beam in beam_outputs]
         corrprod_outputs = self.get_command_arguments(_parse_key_val, "--corrprod")
         corrprod_names = [corrprod["name"] for corrprod in corrprod_outputs]
 
@@ -284,13 +285,15 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
+            default_delays_str = str([0.0] * antennas)
             self.sensors.add(
                 Sensor(
                     str,
                     f"{beam_name}.delay",
-                    "The delay settings of the inputs for this beam: (loadmcnt "
-                    "<ADC sample count when model was loaded>, delay <in seconds>,"
-                    "phase <radians>, ...)",
+                    "The delay settings of the inputs for this beam. Each input has "
+                    "a delay [s] and phase [rad]: (loadmcnt, delay0, phase0, delay1,"
+                    "phase1, ...)",
+                    default=f"(0, {default_delays_str[1:-1]})",
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
@@ -299,7 +302,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     float,
                     f"{beam_name}.quantiser-gain",
                     "Non-complex post-summation quantiser gain applied to this beam",
-                    default=0.0,
+                    default=1.0,
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
@@ -308,6 +311,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     str,
                     f"{beam_name}.weight",
                     "The summing weights applied to all the inputs of this beam",
+                    default=str([1.0] * antennas),
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -285,7 +285,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
-            default_delays_str = str([0.0] * antennas)
+            default_delays = (0,) + (0.0, 0.0) * antennas
             self.sensors.add(
                 Sensor(
                     str,
@@ -293,7 +293,7 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     "The delay settings of the inputs for this beam. Each input has "
                     "a delay [s] and phase [rad]: (loadmcnt, delay0, phase0, delay1,"
                     "phase1, ...)",
-                    default=f"(0, {default_delays_str[1:-1]})",
+                    default=str(default_delays),
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -308,7 +308,6 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     str,
                     f"{beam_name}.weight",
                     "The summing weights applied to all the inputs of this beam",
-                    default="",
                     initial_status=Sensor.Status.NOMINAL,
                 )
             )
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -269,8 +269,59 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
         super().__init__(*args, **kwargs)
         channel_offset = self.get_command_argument(int, "--channel-offset-value")
         channels_per_substream = self.get_command_argument(int, "--channels-per-substream")
+        beam_outpus = self.get_command_arguments(_parse_key_val, "--beam")
+        beam_names = [beam["name"] for beam in beam_outpus]
         corrprod_outputs = self.get_command_arguments(_parse_key_val, "--corrprod")
         corrprod_names = [corrprod["name"] for corrprod in corrprod_outputs]
+
+        for beam_name in beam_names:
+            self.sensors.add(
+                Sensor(
+                    str,
+                    f"{beam_name}.chan-range",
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1357,7 +1399,7 @@ def get_baseline_index(a1, a2):
             if isinstance(stream, product_config.GpucbfBaselineCorrelationProductsStream):
                 renames = ["chan-range", "rx.synchronised", "xeng-clip-cnt"]
             elif isinstance(stream, product_config.GpucbfTiedArrayChannelisedVoltageStream):
-                renames = ["chan-range", "beng-clip-cnt"]
+                renames = ["chan-range", "delay", "quantiser-gain", "weight", "beng-clip-cnt"]
```
bmerry, line 1402: It actually looks like chan-range isn't specified (in the ICD) for B streams, but I feel like that's an oversight. @james-smith-za are you happy if we add it to the ICD?

src/katsdpcontroller/fake_servers.py
```diff
@@ -269,8 +269,58 @@ def __init__(self, *args, **kwargs) -> None:
...
+                    default=f"({channel_offset},{channel_offset + channels_per_substream - 1})",
+                    initial_status=Sensor.Status.NOMINAL,
+                )
+            )
+            self.sensors.add(
+                Sensor(
+                    str,
+                    f"{beam_name}.delay",
+                    "The delay settings of the inputs for this beam: (loadmcnt "
+                    "<ADC sample count when model was loaded>, delay <in seconds>,"
```
james-smith-za, line 292:
```suggestion
                    ", delay [s],"
```

Again, I don't think this first part is necessary.
You can abbreviate the units for the delay, as well as phase as `[rad]` without a loss of information.

src/katsdpcontroller/fake_servers.py
```diff
@@ -269,8 +269,58 @@ def __init__(self, *args, **kwargs) -> None:
...
+                    "phase <radians>, ...)",
+                    initial_status=Sensor.Status.NOMINAL,
```
james-smith-za, line 294: See my comments for the other PR for a sane default. This is the "external" interface so the CAM ICD applies.

Initial delay will be one zero for the loadmcnt, then two zeroes for each antenna (or one for each input).
Probably something like 
```
default=str([0.0]*(n_ants + 1)),
```
may work.

src/katsdpcontroller/fake_servers.py
```diff
@@ -269,8 +269,58 @@ def __init__(self, *args, **kwargs) -> None:
...
+                    "Non-complex post-summation quantiser gain applied to this beam",
+                    default=0.0,
+                    initial_status=Sensor.Status.NOMINAL,
+                )
+            )
+            self.sensors.add(
+                Sensor(
+                    str,
+                    f"{beam_name}.weight",
+                    "The summing weights applied to all the inputs of this beam",
```
james-smith-za, line 310: Default summing weight is 1. So a similar comment to the above, but 1 instead of 0.

src/katsdpcontroller/fake_servers.py
```diff
@@ -294,7 +344,7 @@ def __init__(self, *args, **kwargs) -> None:
                 Sensor(
                     str,
                     f"{corrprod_name}.chan-range",
-                    "The range of channels processed by this XB-engine, inclusive",
+                    "The range of channels processed by this X-engine, inclusive",
```
james-smith-za, line 347: Why this change?

src/katsdpcontroller/generator.py
bmerry: This is going to make sensors called `{stream.name}.{i}.delay` etc, while the ICD specifies just a single sensor called `{stream_name}.delay`.

I think we'll need a new aggregate sensor class for these, which should probably take the child sensor with the latest timestamp.

bmerry (changes requested):

src/katsdpcontroller/fake_servers.py
bmerry, line 1: Need to bump the copyright year.

bmerry (changes requested):

src/katsdpcontroller/generator.py
james-smith-za: These are good suggestions.

src/katsdpcontroller/fake_servers.py
bmerry: While not explicitly stated, I would think loadmcnt ought to be an integer since you can't start applying a delay half-way between samples.

src/katsdpcontroller/generator.py
amishatishpatel: > This is going to make sensors called `{stream.name}.{i}.delay` etc, while the ICD specifies just a single sensor called `{stream_name}.delay`.

Indeed, I was meant to flag that on the ticket earlier when I was going through the full `?sensor-list`, must've got sidetracked. Are you fine if I defer the creation of that new aggregate sensor to the katcp control ticket? Or can I just create a separate ticket for that?

As for the chan-range sensor, I'll make a suggestion on the ICD now.

bmerry: Definitely don't do it in this PR. I've implemented on my `NGC-446-beamformer-control` branch, which builds on top of this one.

amishatishpatel: Noted 👍

src/katsdpcontroller/fake_servers.py
```diff
@@ -284,13 +285,15 @@ def __init__(self, *args, **kwargs) -> None:
...
                 Sensor(
                     str,
                     f"{beam_name}.delay",
-                    "The delay settings of the inputs for this beam: (loadmcnt "
-                    "<ADC sample count when model was loaded>, delay <in seconds>,"
-                    "phase <radians>, ...)",
+                    "The delay settings of the inputs for this beam. Each input has "
+                    "a delay [s] and phase [rad]: (loadmcnt, delay0, phase0, delay1,"
+                    "phase1, ...)",
+                    default=f"(0, {default_delays_str[1:-1]})",
```
bmerry, line 296: This string slicing is a bit icky. I think it's also wrong, because you need a delay AND a phase per antenna, so you need twice as many 0.0's as you have.

Suggestion (untested!):
```python
default_delays = (0,) + (0.0, 0.0) * antennas  # loadmcnt plus delay and phase per antenna
...
        default=str(default_delays),
```

bmerry (changes requested):

src/katsdpcontroller/fake_servers.py
amishatishpatel: Argh, too right. Sure enough, your suggestion does the trick. Please and thank you!

Agreed Re: String slicing. I meant to leave a comment to say we're slicing to get rid of the brackets in the string expansion.

src/katsdpcontroller/generator.py
```diff
@@ -990,6 +990,15 @@ def _make_xbgpu(
         data_suspect_sensors.append(data_suspect_sensor)
 
         stream_sensors: List[Sensor] = [
+            # The timestamps are simply ADC sample counts
+            Sensor(
+                float,
+                f"{stream.name}.scale-factor-timestamp",
+                "Factor by which to divide instrument timestamps to convert to unix seconds",
```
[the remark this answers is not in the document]
amishatishpatel: Argh, I changed the wrong sensor description! (Changed in `make_fgpu` and not here)

src/katsdpcontroller/fake_servers.py
amishatishpatel: This description now matches the ICD exactly (as does the equivalent B-engine sensor).

bmerry (approved):

PR #716, opened 2024-02-20
Upgrade all dependent packages to current versions

bmerry:
This is a fairly substantial change. Apart from the headline of upgrading dependencies, there are a number of related changes:

- I ditched katsdpdockerbase as a source of default dependencies, since most of it hasn't been updated for years. Instead, use pip-compile. I also added a pre-commit hook for `pip-compile`, which will ensure that the requirements files are valid.
- pip-compile is at its best when all the packages are fixed versions (from a package index). I've thus switched the kat* dependencies to their publicly-released versions. katsdptelstate has an unreleased bug fix (for SPR1-2856) which might need to be released before this can be merged.
- pytest 0.23 changes the mechanism for overriding the event loop. While the old way still works, it gives deprecation warnings, so I updated it.
- aiomonitor 0.7.0 opens an extra port with a web UI. It's necessary to ensure this port is allocated via Mesos/Singularity so that two processes don't both try to open the same default port on the same machine.  (See NGC-1179)
- aiohttp 3.9 gives deprecation warnings about using string keys to store data in the app, and prefers using instances of AppKey (in which it mangles the package name into the underlying key to give better uniqueness). This change also gives mypy more visibility into the types being stored, which required some extra type annotation stuff in the tests that do mocking with an object of a different type.

Checklist
- [x] Make a new release of katsdptelstate (and maybe also katdal) and reference it, since there are unreleased changes: https://github.com/ska-sa/katsdptelstate/pull/137
- [x] Run a test in the sandbox
- [x] Run a test with CBF in the lab
- [x] Run a test with SDP in the lab (or on-site, if the lab is non-functional)
- [x] Check the documentation to see if anything needs updating.


bmerry:
@KimMcAlpine requesting a review from you mainly to help with running a test of this, to ensure that I'm not breaking anything in SDP.

Dockerfile
```diff
@@ -23,7 +23,7 @@ ENV PATH="$PATH_PYTHON3" VIRTUAL_ENV="$VIRTUAL_ENV_PYTHON3"
 
 # Install Python dependencies
 COPY requirements.txt /tmp/install/requirements.txt
-RUN install_pinned.py -r /tmp/install/requirements.txt
```
james-smith-za, line 26: Is install_pinned.py something on the docker base image?

bmerry: Yes. It handles an extended form of requirements file that allows another file to be imported as default versions, and it checks that every package has a version pin.

KimMcAlpine:
> @KimMcAlpine requesting a review from you mainly to help with running a test of this, to ensure that I'm not breaking anything in SDP.

I see that the 'Run a test with SDP' checkbox is ticked. Did you have success ? And did you use the lab or site ?

KimMcAlpine:
> > I see that the 'Run a test with SDP' checkbox is ticked. Did you have success ? And did you use the lab or site ?
> 
> I used the lab, and it was successful. I didn't test imaging - as far as I recall, it isn't supposed in the lab? Or am I out of date?

That is correct, no imaging in the lab :)

richarms:
Given that so many other SDP packages (imager, ingest, contim, etc) have been set up to use the `install_pinned.py` script, does it not make more sense to update katsdpdockerbase's base requirements, rather than cutting loose `katsdpcontroller` (for an equivalent number of changes)?

bmerry:
> Given that so many other SDP packages (imager, ingest, contim, etc) have been set up to use the `install_pinned.py` script, does it not make more sense to update katsdpdockerbase's base requirements, rather than cutting loose `katsdpcontroller` (for an equivalent number of changes)?

The problem is that katsdpcontroller is shared with CBF, and with the GPU CBF being in heavy development, we need to update the requirements fairly often. When that happens, we don't want to be blocked on SDP to update katsdpdockerbase, which is a significant undertaking given the number of packages that are affected by such an update.

requirements.txt
```diff
@@ -1,53 +1,318 @@
...
+    # via
+    #   aiohttp
+    #   katsdpcontroller (setup.cfg)
+zipp==3.17.0
+    # via
+    #   importlib-metadata
+    #   importlib-resources
 
-katpoint @ git+https://github.com/ska-sa/katpoint
-katdal @ git+https://github.com/ska-sa/katdal
```
ludwigschwardt, line 50: Do you lose the ability to run off GitHub? That might require us to release more often to satisfy katsdpcontroller (although I doubt it will actually happen 😅).

richarms:
> > Given that so many other SDP packages (imager, ingest, contim, etc) have been set up to use the `install_pinned.py` script, does it not make more sense to update katsdpdockerbase's base requirements, rather than cutting loose `katsdpcontroller` (for an equivalent number of changes)?
> 
> The problem is that katsdpcontroller is shared with CBF, and with the GPU CBF being in heavy development, we need to update the requirements fairly often. When that happens, we don't want to be blocked on SDP to update katsdpdockerbase, which is a significant undertaking given the number of packages that are affected by such an update.

This isn't a new problem, where one package is under more active development than its brothers, and is in fact one of the main reasons that `katsdpdockerbase` is so out of date -- it's undesirable to trigger downstream dependency issues. 

At some point (I believe!) the logic went that it was better to have consistent package lists *somewhere* than to not have to deal with downstream hell. This comment (and this commit) perhaps marks the end of that line of thinking.

Dockerfile
ludwigschwardt: It's totally sweet and awesome Merriness. Its pinning is better than Poetry :-P

requirements.txt
bmerry: I think it might work okay if we pin to a specific commit, but I've had some poor experiences in the past with mixing Git URLs with pip-compile.

Just embrace release-early, release-often @ludwigschwardt  :-)

bmerry:
@KimMcAlpine the checklist is all done. Do you want anyone in Pipelines to review the changes before merging?

KimMcAlpine:
> @KimMcAlpine the checklist is all done. Do you want anyone in Pipelines to review the changes before merging?
I looked at it yesterday. I will approve it now. I don't think we need another person from pipelines to review it

KimMcAlpine (approved):

bmerry:
Thanks @KimMcAlpine. I'll merge. Be aware that next time you deploy to site, you'll need to update the master controller as well. It might seem like it's working if you don't, but when you do subarrays and end up with two product controllers running on the same machine you'll encounter problems.

PR #719, opened 2024-02-20
Fix errors from dash when product controller not yet built

bmerry:
When a612045 added an extra return value to `top_level`, not all the return paths were updated to return it.

Additionally, fix up a comment that black moved to the wrong place.

### DIFF
--- a/src/katsdpcontroller/dashboard.py
+++ b/src/katsdpcontroller/dashboard.py
@@ -279,14 +279,14 @@ class Dashboard:
                 Output("task-table", "data"),
                 Output("capture-block-table", "data"),
                 Output("batch-table", "data"),
-                Output("root", "data-dummy"),
-            ],  # Just to set loading state on root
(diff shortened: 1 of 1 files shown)


richarms:
I re-started the build on Jenkins after removing about 20 ghost nodes, and it now succeeds.

src/katsdpcontroller/dashboard.py
```diff
@@ -279,14 +279,14 @@ def _make_app(self, **dash_args):
                 Output("task-table", "data"),
                 Output("capture-block-table", "data"),
                 Output("batch-table", "data"),
-                Output("root", "data-dummy"),
-            ],  # Just to set loading state on root
+                Output("root", "data-dummy"),  # Just to set loading state on root
+            ],
```
richarms, line 283: Hard to say whether this will in general change the semantic association of the comment, but pretty poor form by `black`!

richarms (approved):

PR #720, opened 2024-02-21
Make sandbox testing work again

bmerry:
In #717 I was too clever and merged the `lo` interface into a single katsdpcontroller interface attached to three networks. It turns out that functionality isn't completely functional (SPR1-3043) so split it back out again.


james-smith-za (approved):
Nice sleuthing.

PR #723, opened 2024-03-14
Spr1 2999 set cap net raw on bfingest

KimMcAlpine:
Enable ibverbs in bfingest


bmerry (approved):
Ok yes, I remember Ludwig discovering he needed to make this change.

PR #724, opened 2024-03-25
Update to the latest version of async_solipsism

bmerry:
It now provides its own EventLoopPolicy, so we don't need to write one in test/utils.py.


amishatishpatel (approved):
Excellent.

PR #727, opened 2024-04-15
Add GPU resource allocations for xbgpu with beams

bmerry:
The compute usage is just a guess based on being able to do 64 beams in tests. The memory usage appears accurate from tests though.

Closes NGC-1222.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1107,6 +1107,7 @@ def _make_xbgpu(
     batches_per_chunk = math.ceil(target_chunk_size / batch_size)
     chunk_size = batches_per_chunk * batch_size
     rx_reorder_tol = 2**29  # Default of --rx-reorder-tol option
+    n_tx_items = 2  # Magic number default for xbgpu
     chunk_ticks = acv.n_samples_between_spectra * acv.n_spectra_per_heap * batches_per_chunk
     max_active_chunks = math.ceil(rx_reorder_tol / chunk_ticks) + 1  # Based on calc in xbgpu
     free_chunks = max_active_chunks + 8  # Magic number is from xbgpu
@@ -1168,7 +1169,7 @@ def _make_xbgpu(
                 mid_vis_size = max(100 * 1024 * 1024, vis_size * 2)
                 xbgpu.mem += _mb(vis_size * 5)  # Magic number is default in XSend class
                 xbgpu.gpus[0].compute += 0.15 * bw_scale
-                xbgpu.gpus[0].mem += _mb(2 * vis_size + mid_vis_size)
+                xbgpu.gpus[0].mem += _mb(n_tx_items * vis_size + mid_vis_size)
                 # Minimum capability as a function of bits-per-sample, based on
                 # tensor_core_correlation_kernel.mako from katgpucbf.xbgpu.
                 min_compute_capability = {4: (7, 3), 8: (7, 2), 16: (7, 0)}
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1166,20 +1168,23 @@ def get_baseline_index(a1, a2):
...
+                elements = (
+                    batches_per_chunk * stream.n_chans_per_substream * stream.spectra_per_heap
                 )
-                xbgpu.mem += _mb(beam_size * 2)  # Magic number default for xbgpu
+                beam_size = elements * COMPLEX
+                rand_state_size = elements * 48  # 48 is sizeof(curandStateXORWOW_t)
+                xbgpu.mem += _mb(beam_size * n_tx_items)
+                # Allow 128 single-pol beams + 1 baseline-correlation-products for 80 antennas.
+                xbgpu.gpus[0].compute += 0.125 / n_inputs * bw_scale
+                xbgpu.gpus[0].mem += _mb(beam_size * n_tx_items)
```
amishatishpatel, line 1186: If I could be a small, but non-zero pain
* Could I request you update this calc to be in the ~same order as the one for X-engine config?
* i.e. `_mb(n_tx_items * beam_size)`  - "multiplier" first

src/katsdpcontroller/generator.py
```diff
@@ -1166,20 +1168,23 @@ def get_baseline_index(a1, a2):
...
-                    * COMPLEX
+                elements = (
+                    batches_per_chunk * stream.n_chans_per_substream * stream.spectra_per_heap
                 )
-                xbgpu.mem += _mb(beam_size * 2)  # Magic number default for xbgpu
+                beam_size = elements * COMPLEX
+                rand_state_size = elements * 48  # 48 is sizeof(curandStateXORWOW_t)
+                xbgpu.mem += _mb(beam_size * n_tx_items)
+                # Allow 128 single-pol beams + 1 baseline-correlation-products for 80 antennas.
+                xbgpu.gpus[0].compute += 0.125 / n_inputs * bw_scale
```
amishatishpatel, line 1185: Not to bluntly say "I don't see how the comment and the line inform each other", but I will take your word that one ~dictated the other.

src/katsdpcontroller/generator.py
```diff
@@ -1166,20 +1168,23 @@ def get_baseline_index(a1, a2):
...
-                    batches_per_chunk
-                    * stream.n_chans_per_substream
-                    * stream.spectra_per_heap
-                    * COMPLEX
+                elements = (
+                    batches_per_chunk * stream.n_chans_per_substream * stream.spectra_per_heap
                 )
-                xbgpu.mem += _mb(beam_size * 2)  # Magic number default for xbgpu
+                beam_size = elements * COMPLEX
+                rand_state_size = elements * 48  # 48 is sizeof(curandStateXORWOW_t)
```
amishatishpatel, line 1182: Appreciate the comment

src/katsdpcontroller/generator.py
```diff
@@ -1146,10 +1147,11 @@ def get_baseline_index(a1, a2):
...
         )
         xbgpu.gpus = [scheduler.GPURequest()]
-        xbgpu.gpus[0].compute = (
-            0.15 * bw_scale
-        )  # TODO: NGC-1222 update depending on number and type of streams
         xbgpu.gpus[0].mem = 300 + _mb(3 * chunk_size)
+        # This memory is independent of the number of beams, but is needed
+        # only if there is at least one beam. So set it to 0 here then
+        # update it if we see a beam.
+        rand_state_size = 0
```
amishatishpatel, line 1154: Is this currently true? Or does it depend on the update in https://github.com/ska-sa/katgpucbf/pull/731?

Either way, noted.

amishatishpatel (approved):

amishatishpatel:
Apologies, I hit Approve when I should've said "Request Changes".

amishatishpatel (changes requested):
Apologies, I hit Approve when I should've said "Request Changes".

src/katsdpcontroller/generator.py
bmerry: Done.

src/katsdpcontroller/generator.py
bmerry: If you repeat this line 128 times and there are 160 inputs (80 ants) and bw_scale=1 (worst case) then this comes to 0.1, which when added to the 0.15 for the X pipeline gives 0.25 i.e. 4 engines per GPU as desired.

amishatishpatel (approved):

PR #728, opened 2024-04-16
Reduce logging for large katcp messages

bmerry:
issue_req logs the full message that is being issued to the task, and again when it is successful. For large messages like ?gain, this can take a significant portion of the message processing time.

Prune the message by limiting the logging to 20 arguments.

See NGC-1233.

### DIFF
--- a/src/katsdpcontroller/tasks.py
+++ b/src/katsdpcontroller/tasks.py
@@ -452,7 +452,9 @@ class ProductPhysicalTaskMixin(scheduler.PhysicalNode):
                 f"Cannot issue request {req} to node {self.name} without a katcp connection"
             )
         log_args = args
-        # Don't bother trimming if it won't be logged anyway
+        # Limit the number of arguments we report to avoid spamming the logs
+        # with huge messages. But don't bother trimming if it won't be logged
+        # anyway.
         if self.logger.isEnabledFor(log_level):
             max_args = 20
             if len(args) > max_args:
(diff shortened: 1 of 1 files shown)


james-smith-za (approved):
This change is probably necessary but I'm slightly uncomfortable on two points.

The first is that we won't be able to see in the logs if something is wrong with arg[20:]. I guess the risk of this is low: we'd affect some funkiness to probably show up in the first twenty arguments, not only later ones.

The second is that the reason for this clipping is not entirely obvious to the reader. The astute investigator could probably look through the `blame` for lines 444-449, but I'd feel more warm and fuzzy inside with a one-liner at the top explaining that we are capping the number of arguments logged at 20 for performance reasons.

This is bike-shedding though so at your discretion.

bmerry:
> The first is that we won't be able to see in the logs if something is wrong with arg[20:]. I guess the risk of this is low: we'd affect some funkiness to probably show up in the first twenty arguments, not only later ones.

Logging is not the right place for this sort of bulk data anyway - next thing you'll be wanting to log all the visibilities in case they're wrong. The many-argument requests I can think of set values that are reported as sensors, so it's not completely opaque.

> The second is that the reason for this clipping is not entirely obvious to the reader.

I've added a comment. I also mis-interpreted your comment originally and thought you meant the reader of the logs, so I changed things a bit so that the log message is explicit that arguments were omitted and how many.

james-smith-za (approved):
Fair points.

I like the changes, yes I was talking about the reader of the code but letting the reader of the logs know that they've been truncated (or that some have been elided as you describe it) is also a good idea.

PR #730, opened 2024-04-17
Strip out support for product-configure schema version 2

bmerry:
We've been using version 3 for many years, and I want to get rid of version 2 before adding version 4 to keep the code reasonably maintainable.


bmerry:
@ludwigschwardt @KimMcAlpine @richarms FYI. I've asked @james-smith-za to do the review.

ludwigschwardt:
Nah, when it comes to the live system, I can live on the edge. 😁  It's the old datasets that worry me.

PR #731, opened 2024-04-18
Improvements to sync epoch setting and querying

bmerry:
NB: this needs to be merged together with https://github.com/ska-sa/katgpucbf/pull/737. Without it, things will break.

This implements versions 4 of the product-configure schema (see the [changelog](https://docs.google.com/document/d/15ZIyuBf4Vk4ESIEdKsQxdSBoLEwh85eyKeDJouLRAZY/edit#heading=h.bxv6wohps6p4) and SPR1-3072). It also adds support for sync-time sensors exported from the dsims (coming in that mentioned PR).

### DIFF
--- a/src/katsdpcontroller/schemas/__init__.py
+++ b/src/katsdpcontroller/schemas/__init__.py
@@ -17,6 +17,7 @@
 """Makes packaged JSON schemas available."""
 
 import json
+from typing import TYPE_CHECKING, Union
 
 import importlib_resources
 import jinja2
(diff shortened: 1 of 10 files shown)


src/katsdpcontroller/schemas/__init__.py
```diff
@@ -26,12 +27,12 @@
 _env = jinja2.Environment(loader=jinja2.PackageLoader(__name__, "."))
 
 
-def _normalise_version(version):
+def _normalise_version(version: Union["ComparableVersion", str]) -> "ComparableVersion":
```
james-smith-za, line 30: Can we not use the `|` syntax? Or are we still stuck on Python 3.8 for the controller?

james-smith-za (approved):
Seems rational.

src/katsdpcontroller/schemas/__init__.py
bmerry: The controller still uses the katsdpdockerbase base image, which is Python 3.8.

PR #739, opened 2024-06-05
Add shutdown_delay to the "config" dict in product-configure

bmerry:
This allows the default delay of 10s for be overridden, which is useful for NGC-1343.


james-smith-za (approved):
Seems fine.

PR #740, opened 2024-06-05
Remove depends_ready links for fgpu/xbgpu

bmerry:
This allows the containers for dsim, fgpu and xbgpu to all start at the same time, instead of in phases, which should speed up correlator startup. This is safe because none of the processes communicates with the prior ones other than via the multicast groups.

The depends_init links are left in place. They don't affect operation (they're used by exec_transitions, which isn't applicable to CBF), but I modified the dashboard to take them into account for sorting purposes. This was easier than introducing yet another property and applying it all over the place.

See NGC-1343.


bmerry:
It's hard to say exactly how much time this saves because the config time is quite variable, but it seems to be at least 10s saving.

james-smith-za (approved):
Let's see how much of a difference these make on qualification overnight.

PR #745, opened 2024-06-18
{src, dst} -> {recv, send} for GPU CBF

amishatishpatel:
Contributes to: NGC-1158.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -693,11 +693,11 @@ def _make_fgpu(
             + taskset
             + [
                 "fgpu",
-                "--src-interface",
+                "--recv-interface",
                 "{interfaces[gpucbf].name}",
-                "--dst-interface",
+                "--send-interface",
                 "{interfaces[gpucbf].name}",
-                "--dst-packet-payload",
+                "--send-packet-payload",
                 "8192",
                 "--adc-sample-rate",
                 str(srcs[0].adc_sample_rate),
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -742,26 +742,26 @@ def _make_fgpu(
 
         if not configuration.options.develop.less_resources:
             fgpu.command += [
-                "--src-affinity",
+                "--recv-affinity",
                 "{cores[src]}",
```
bmerry, line 746: I think rename the cores as well (line 664).

src/katsdpcontroller/generator.py
```diff
@@ -1251,22 +1251,22 @@ def get_baseline_index(a1, a2):
                 ]
 
         if not configuration.options.develop.less_resources:
-            xbgpu.command += ["--src-affinity", "{cores[src]}", "--dst-affinity", "{cores[dst]}"]
+            xbgpu.command += ["--recv-affinity", "{cores[src]}", "--send-affinity", "{cores[dst]}"]
```
bmerry, line 1254: Again, rename the cores (line 1123).

bmerry (changes requested):
Looks reasonable to me - obviously will need to be matched with a katgpucbf PR. I've got one suggestion to extend the change a little further into the katsdpcontroller code.

amishatishpatel:
Thanks for being so quick on this, and with an excellent suggestion too.

Corresponding katgpucbf PR is up for review.

james-smith-za (approved):
LGTM.


@bmerry made a good suggestion so IMO once you've incorporated that and the tests pass, I'd say go ahead and merge.

bmerry (approved):

PR #748, opened 2024-06-20
Raise ValueError if LogicalTask.numa_nodes is out of range

bmerry:
A previous bug that left it greater than 1.0 prevented allocating a task anywhere, but with a confusing error message. This will help catch any future such bugs faster.

Closes NGC-1353.

### DIFF
--- a/src/katsdpcontroller/scheduler.py
+++ b/src/katsdpcontroller/scheduler.py
@@ -1606,6 +1606,16 @@ class LogicalTask(LogicalNode, ResourceRequestsContainer):
         self.taskinfo.command.shell = False
         self.physical_factory = PhysicalTask
 
+    @property
+    def numa_nodes(self) -> float:
+        return self._numa_nodes
+
+    @numa_nodes.setter
+    def numa_nodes(self, value: float) -> None:
+        if not 0.0 <= value <= 1.0:
+            raise ValueError("numa_nodes must be between 0 and 1")
+        self._numa_nodes = value
+
     def valid_agent(self, agent) -> bool:
         """Checks whether the attributes of an agent are suitable for running
         this task. Subclasses may override this to enforce constraints e.g.,
--- a/test/test_scheduler.py
+++ b/test/test_scheduler.py
@@ -18,6 +18,7 @@ import asyncio
 import base64
 import ipaddress
 import logging
+import math
 import socket
 import time
 from collections import Counter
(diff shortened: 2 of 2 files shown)


src/katsdpcontroller/scheduler.py
```diff
@@ -1606,6 +1606,16 @@ def __init__(self, name):
...
         self.physical_factory = PhysicalTask
 
+    @property
+    def numa_nodes(self) -> float:
+        return self._numa_nodes
+
+    @numa_nodes.setter
+    def numa_nodes(self, value: float) -> None:
+        if not 0.0 <= value <= 1.0:
+            raise ValueError("numa_nodes must be between 0 and 1")
```
amishatishpatel, line 1616: I'm not even sure this question is warranted, so apologies for making you answer it. 
* Does this error message accurately describe the range of valid/required values for `numa_nodes`?
* i.e. Does it provide enough information that it can be a fractional/float value?

An error message that reads "... must be between 0.0 and 1.0" looks a bit odd
* I guess anyone using altering this can see the type-hints in this base class, and
* That should be enough ?

Let me know what your thinking was/is. Please and thank you.

test/test_scheduler.py
```diff
@@ -1250,6 +1251,24 @@ def test_allocate_success(self):
...
+        assert task.numa_nodes == 0.0
+        task.numa_nodes = 0.5
+        assert task.numa_nodes == 0.5
+        with pytest.raises(ValueError):
+            task.numa_nodes = 1.2
+        with pytest.raises(ValueError):
+            task.numa_nodes = -0.1
+        with pytest.raises(ValueError):
+            task.numa_nodes = math.nan
+        assert task.numa_nodes == 0.5  # Must not have been altered by failures
```
amishatishpatel, line 1269: A small part of me thought "Why's he testing if the value is 0.5 twice?", but really that's the whole point of this addition/update.
* Need to check the initial update works correctly,
* Then check any erroneous updates/attempts to update don't affect it.

amishatishpatel (commented):
Just one comment before I explicitly Approve. Thanks in advance for humouring me.

src/katsdpcontroller/scheduler.py
bmerry: The docstring says
```
    numa_nodes : float
        Minimum fraction of a NUMA node's cores to allocate. This should only
        be set if `cores` is, and must not be set higher than 1.0.
```
which seems pretty clear that it's a fraction.

test/test_scheduler.py
bmerry: Is there a change you're requesting here?

amishatishpatel: No change requested, just thinking out aloud 😅

amishatishpatel (approved):

PR #749, opened 2024-06-20
Improve diagnose_insufficient

bmerry:
If there is a task for which there is no valid agent, report this explicitly rather than indicating that there were insufficient (0) of some resource.

Also special-case tasks with a pinned host to give a more helpful error.

Closes NGC-1350.


bmerry:
Tested by draining cbfgpu001 and trying to pin a task to it:

Before:
```
Not enough cpus for f.antenna-channelised-voltage.0 on any agent (3.000 > 0.000)
```
After:
```
Task f.antenna-channelised-voltage.0 needs host cbfgpu001.cbf.mkat.karoo.kat.ac.za but no offers were received for it
```

james-smith-za (approved):

PR #750, opened 2024-07-02
Add payload-len sensor for GPU CBF

amishatishpatel:
See ticket for further detail/discussion. This is as per Version 7 of the MK CBF-CAM ICD (M1000-0001-002),
tweaked to conform to the MK+ CBF sensor name convention.

I still have an outstanding question on the ticket regarding whether I/we should explicitly state
the value is in **bytes**.

A screenshot below of the `?sensor-value` queries for a 4A correlator with 4 `--beams`.
image

Resolves: NGC-962.


bmerry (changes requested):
Looks good! Just one formatting request - no need for a re-review when you fix that.

PR #751, opened 2024-07-02
Fix a string that was supposed to be an f-string

bmerry:


amishatishpatel (approved):
Oh dear, where did this come from?

bmerry:
> Oh dear, where did this come from?

I noticed it when running `?sensor-list`.

src/katsdpcontroller/tasks.py
```diff
@@ -430,7 +430,7 @@ def __init__(
             self._interface_sensors[request.network] = Sensor(
                 str,
                 self.name + f".interfaces.{request.network}.name",
```
ludwigschwardt, line 432: I like the mix of the new and the old :-P

amishatishpatel: Dammit I didn't even notice the line above 😅

PR #754, opened 2024-07-12
Rename xengs-synchronised to rx-synchronised

james-smith-za:
This sensor was originally introduced in 02ba772 and renamed from synchronised to xengs-synchronised in 450d34a. Both commits were attributed to NGC-665.

I can't quite see the logic for how we initially settled on this name but I suspect it was in honour of the `xeng-vaccs-synchronised` sensor present on MeerKAT.

The implementation doesn't currently agree with the ICD, which wants `<xeng_stream_name>.rx.synchronised`. This commit fixes that.

Closes NGC-1390

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1013,7 +1013,7 @@ def _make_xbgpu(
                 ),
                 SyncSensor(
                     sensors,
-                    f"{stream.name}.xengs-synchronised",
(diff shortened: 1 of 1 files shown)


amishatishpatel:
Have you run an X-engine/correlator with this change? Just want to see what the namings look like, and if the `sensor_renames` don't affect this.

amishatishpatel (approved):

PR #755, opened 2024-07-12
Send inform count with ?capture-list reply

james-smith-za:
KATCP convention is, in response to a request that results in informs being sent, to send the count of informs with the reply.

Previously this was not done due to the MK+ CBF / CAM ICD not calling for it. We have since decided to update the ICD and follow the convention.

Contributes to NGC-1391.


amishatishpatel:
Could I ask you to paste a screenshot of the updated request, please? Slash, how have you confirmed it has the desired effect?

james-smith-za:
image

amishatishpatel (approved):

PR #759, opened 2024-08-14
Suppress interface-changed while CONFIGURING

bmerry:
This should in theory reduce the thrashing of the master controller asking for a new copy of the sensor list every time a new engine comes online.


bmerry:
@KimMcAlpine is going to organise time to test this (for SDP) on site next Wednesday.

james-smith-za (approved):

PR #761, opened 2024-08-19
Bump aiohttp to the latest version

bmerry:
To keep dependabot happy


bmerry:
Seems like the new version of aiohttp doesn't play nice with async_solipsism. I'm going to turn this into a draft PR (to keep it on my radar) until I have time to look into that.

james-smith-za (approved):

PR #762, opened 2024-08-19
Update Aggregate Sensor Sampling Strategy

amishatishpatel:
Something we noticed when debugging the greater NGC-938,
* even though the constituent sensors in the Engines were obeying their sampling intervals,
* the stream-level AggregateSensor was updating as often as each constituent sensor update. 

This update aims to alleviate some of the sensor update spam, and uses the same update interval 
from katgpucbf.

Contributes to: NGC-938.

### DIFF
--- a/src/katsdpcontroller/defaults.py
+++ b/src/katsdpcontroller/defaults.py
@@ -51,6 +51,9 @@ PFB_TAPS = 16
 #: Default payload size for gpucbf data products. Bigger is better to
 #: minimise the number of packets/second to process.
 GPUCBF_PACKET_PAYLOAD_BYTES = 8192
+#: Minimum update period (in seconds) for katcp sensors where the underlying
+#: value may update extremely rapidly.
+GPUCBF_MIN_SENSOR_UPDATE_PERIOD = 5.0
 #: Autotune fallback behaviour: "nearest" or "exact"
 KATSDPSIGPROC_TUNE_MATCH = "nearest"
 #: Time (in seconds) to sleep before exiting
--- a/src/katsdpcontroller/defaults.py
+++ b/src/katsdpcontroller/defaults.py
@@ -53,7 +53,7 @@ PFB_TAPS = 16
 GPUCBF_PACKET_PAYLOAD_BYTES = 8192
 #: Minimum update period (in seconds) for katcp sensors where the underlying
 #: value may update extremely rapidly.
-GPUCBF_MIN_SENSOR_UPDATE_PERIOD = 5.0
+GPUCBF_MIN_SENSOR_UPDATE_PERIOD = 1.0
 #: Autotune fallback behaviour: "nearest" or "exact"
 KATSDPSIGPROC_TUNE_MATCH = "nearest"
 #: Time (in seconds) to sleep before exiting
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -48,14 +48,14 @@ import katsdpmodels.fetch.aiohttp
 import katsdptelstate.aio
 import networkx
 import numpy as np
-from aiokatcp import Sensor, SensorSet
+from aiokatcp import Sensor, SensorSampler, SensorSet
 from katsdpmodels.band_mask import BandMask, SpectralWindow
 from katsdpmodels.rfi_mask import RFIMask
 from katsdptelstate.endpoint import Endpoint
 
 from . import defaults, product_config, scheduler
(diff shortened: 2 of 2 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1057,29 +1061,38 @@ def get_baseline_index(a1, a2):
                     sensors,
                     f"{stream.name}.beng-clip-cnt",
                     "Number of complex samples that saturated",
+                    # Update this to honour sampling intervals and strategies
```
bmerry, line 1064: Not sure what this comment means.

src/katsdpcontroller/defaults.py
```diff
@@ -51,6 +51,9 @@
 #: Default payload size for gpucbf data products. Bigger is better to
 #: minimise the number of packets/second to process.
 GPUCBF_PACKET_PAYLOAD_BYTES = 8192
+#: Minimum update period (in seconds) for katcp sensors where the underlying
+#: value may update extremely rapidly.
+GPUCBF_MIN_SENSOR_UPDATE_PERIOD = 1.0
```
bmerry, line 56: I think let's call it something like `FAST_SENSOR_UPDATE_PERIOD`. We're applying it because it updates too quickly otherwise, not because it's specific to gpucbf.

bmerry:
I'm feeling slightly nervous about the range of sensors we're applying this logic to. For xeng-clip-cnt and beng-clip-cnt I'm not worried, but some of the other sensors are used in conjunction with the data (particularly around delays). @james-smith-za @ludwigschwardt any thoughts?

For LatestSensor, we might consider modifying the class implementation so that it suppresses updates that only uptime the timestamp by a small amount without updating the value. That'll help to combine the storm of duplicated updates from all the engines, while still immediately passing on actual changes.

src/katsdpcontroller/generator.py
```diff
@@ -1010,6 +1010,8 @@ def get_baseline_index(a1, a2):
                     "Number of visibilities that saturated",
                     name_regex=re.compile(rf"{escaped_name}\.[0-9]+\.xeng-clip-cnt"),
                     n_children=stream.n_substreams,
+                    auto_strategy=SensorSampler.Strategy.EVENT_RATE,
+                    auto_strategy_parameters=(GPUCBF_MIN_SENSOR_UPDATE_PERIOD, math.inf),
```
bmerry, line 1014: To reduce duplicated code, I think it'll work to have a module-level constant

```python
FAST_SENSOR_UPDATE_PARAMS = {
    "auto_strategy": SensorSampler.Strategy.EVENT_RATE,
    "auto_strategy_parameters": (GPUCBF_MIN_SENSOR_UPDATE_PERIOD, math.inf),
}
```
and then here you could do
```suggestion
                    **FAST_SENSOR_UPDATE_PARAMS,
```

You'll just have to see what mypy thinks of it though. I think mypy complaints might be why I didn't use this pattern in katgpucbf, although it might also be a frog-in-boiling-water effect (initially there were only a couple of usages so it wasn't worth the effect, then applying it to one new sensor was always easier than doing the refactoring).

bmerry (changes requested):
Feel free to skip my complicated plan to de-duplicate code. It's a idea that seemed worth writing down rather than something I think needs fixing.

ludwigschwardt:
I'm a bit fuzzy exactly how event-rate sampling works, but as long as the delay *value* timestamps are correct (and we don't lose any because of this scheme), then I'm happy.

Also, what does an aggregate delay sensor even mean? How does it combine delays?

bmerry:
> I'm a bit fuzzy exactly how event-rate sampling works

After sending an update, it suppresses sending more updates for a certain time. If there further changes during that window, it'll send the last reading at the end of the window.

, but as long as the delay value timestamps are correct (and we don't lose any because of this scheme), then I'm happy.

The timestamps should be correct, but you might lose some updates. I think that only actually occurs if you schedule delays with load times that are closer together than the event-rate min period.

>Also, what does an aggregate delay sensor even mean? How does it combine delays?

It just takes the latest reading (highest timestamp) as the aggregate. Nominally the engines should all be reporting the same delays, but the `?beam-delays` request doesn't take a load timestamp so the individual engines may receive and action the request at slightly different times.

amishatishpatel:
Apologies, I didn't meant to close this PR (and delete the branch). I tried to do something unnecessary and rename the branch to correspond to the new/spin-off ticket (NGC-1459). Should've known it would have unintended side-effects.

I'll raise another PR with the comments noted (see the tickets for more/any updates). Apologies once more!

PR #763, opened 2024-08-21
Set a lower bound on spectra per chunk for xbgpu

bmerry:
This will increase memory usage somewhat, but is needed to get xbgpu to
perform at satisfactory levels.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1124,7 +1124,10 @@ def _make_xbgpu(
         // 8
     )
     target_chunk_size = 64 * 1024**2
-    batches_per_chunk = math.ceil(target_chunk_size / batch_size)
+    # Performance is poor if the number of spectra per chunk is too low,
+    # so we ensure at least 128 even if that means overshooting the
+    # target chunk size.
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1199,7 +1202,7 @@ def get_baseline_index(a1, a2):
                     batches_per_chunk * stream.n_chans_per_substream * stream.spectra_per_heap
                 )
                 beam_size = elements * COMPLEX
-                rand_state_size = elements * 48  # 48 is sizeof(curandStateXORWOW_t)
+                rand_state_size = elements * 24  # 24 is sizeof(randState_t)
```
amishatishpatel, line 1205: Is this to accommodate for the increase in GPU memory usage? (Line 1330)

Doesn't necessarily seem related to NGC-1457.

amishatishpatel (approved):
Just the one comment on your update to `rand_state_size`.

src/katsdpcontroller/generator.py
amishatishpatel: Apologies, I see the second commit corresponds to this, and the commit message explains it.

bmerry: Indeed, I probably ought to have put it in a separate PR.

PR #765, opened 2024-08-22
Add GPU CBF B-engine telstate info

amishatishpatel:
As mentioned on the ticket, I haven't added the `subarray_product_id` key/value in here.
1. Because I don't yet know how, but
2. Because it seems it wasn't mandatory for CBF to populate that.

Whenever you can, please and thank you.

Resolves: NGC-1225.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1104,7 +1104,15 @@ def _make_xbgpu(
                 n_accs=n_accs,
             )
         elif isinstance(stream, product_config.GpucbfTiedArrayChannelisedVoltageStream):
-            # TODO: NGC-1225
+            telstate_data.update(
+                # Had to keep mismatch in spelling as external users still use
+                # 'cent-e-r' spelling.
+                center_freq=stream.centre_frequency,
+                n_chans=stream.n_chans,
+                scale_factor_timestamp=stream.adc_sample_rate,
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1104,7 +1104,15 @@ def get_baseline_index(a1, a2):
...
-            # TODO: NGC-1225
+            telstate_data.update(
+                # Had to keep mismatch in spelling as external users still use
+                # 'cent-e-r' spelling.
+                center_freq=stream.centre_frequency,
+                n_chans=stream.n_chans,
+                scale_factor_timestamp=stream.adc_sample_rate,
+                sync_time=sync_time,
+                ticks_between_spectra=stream.n_samples_between_spectra,
+            )
```
bmerry, line 1115: See [this comment](https://skaafrica.atlassian.net/browse/NGC-1225?focusedCommentId=223640): we don't need any of these (SDP gets them from the upstream antenna-channelised-voltage stream), but we do need spectra_per_heap.

bmerry (changes requested):

src/katsdpcontroller/generator.py
amishatishpatel: My apologies once more. Especially because spectra_per_heap is the only one I left out 🤦

PR #768, opened 2024-09-05
Bump aiokatcp version to 2.0.0

amishatishpatel:
This version bump does result in a breaking change for setting sensor values.
Hence, the tweaks to unit tests and some sensors.

Contributes to: NGC-1062.

### DIFF
--- a/test/test_web.py
+++ b/test/test_web.py
@@ -308,7 +308,7 @@ class TestWeb:
 
     async def test_update_bad_gui_sensor(self, mc_server, caplog) -> None:
         with caplog.at_level(logging.ERROR, logger="katsdpcontroller.web"):
-            mc_server.sensors["product1.gui-urls"].value = "not valid json"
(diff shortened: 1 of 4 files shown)


test/test_web.py
```diff
@@ -218,38 +218,38 @@ def mc_server(self, use_haproxy: bool) -> mock.MagicMock:
         )
         mc_server.sensors.add(
             Sensor(
-                str,
+                bytes,
                 "gui-urls",
                 "",
-                default=json.dumps(ROOT_GUI_URLS),
+                default=json.dumps(ROOT_GUI_URLS).encode(),
                 initial_status=Sensor.Status.NOMINAL,
```
bmerry, line 225: I think this one should stay as `str`, because it's created directly in the master controller server (as `str`). The productN.gui-urls are created by SensorWatcher and so will have type `bytes`.

bmerry (approved):

PR #770, opened 2024-09-26
use harbor as source for base-build

richarms:
Is there a good reason why controller has not been using harbor as the source for base-build?


bmerry:
> Is there a good reason why controller has not been using harbor as the source for base-build?

Yes. Doing so would prevent collaborators outside the SARAO firewall from being able to build things.

PR #772, opened 2024-10-09
Send latest steady-state-timestamp with capture-start

bmerry:
This implements the ICD guarantee that a request that changes processing (e.g. ?gains) followed by ?capture-start will not transmit data that does not yet account for the change.

See NGC-1462.

### DIFF
--- a/test/test_product_controller.py
+++ b/test/test_product_controller.py
@@ -1989,15 +1989,26 @@ class TestController(BaseTestController):
             "delays", "gpucbf_antenna_channelised_voltage", 1234567890.0, "0,1:0,0", "0,1:0,1"
         )
 
-    async def test_capture_start(self, client: aiokatcp.Client, sensor_proxy_client) -> None:
+    async def test_capture_start(
+        self, client: aiokatcp.Client, server: DeviceServer, sensor_proxy_client
+    ) -> None:
         """Test capture-start in the success case."""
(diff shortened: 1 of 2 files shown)


test/test_product_controller.py
```diff
@@ -1989,15 +1989,26 @@ async def test_delays_success(self, client: aiokatcp.Client, sensor_proxy_client
...
+                initial_status=Sensor.Status.NOMINAL,
+            )
+        )
         await client.request("capture-start", "gpucbf_baseline_correlation_products")
         katcp_client = sensor_proxy_client
         # TODO: this doesn't check that the requests are going to the correct
         # nodes.
         katcp_client.request.assert_any_call(
-            "capture-start", "gpucbf_baseline_correlation_products"
+            "capture-start", "gpucbf_baseline_correlation_products", 12345
```
amishatishpatel, line 2011: I love how such a small addition can so properly test the logic you've updated. At least, it made me triple-check I understood the new `?capture-start` path.

amishatishpatel (approved):
Is it possible/useful to test this on a live correlator?

bmerry:
> Is it possible/useful to test this on a live correlator?

I've already done that - just checking a tcpdump of the protocol to ensure that it's passing the value along. But I think a qualification test will be needed to properly verify that this achieves the desired guarantee.

PR #775, opened 2024-10-21
Give dsim more memory

bmerry:
The change to use FFT for MultiCW (ska-sa/katgpucbf#875) increases memory usage.


bmerry:
Each dsim is allocated 4 cores (1/4 of the machine) and 8GB is only 1/8 of the RAM, so I don't think this should have any negative impact on utilisation.

james-smith-za (approved):

PR #779, opened 2024-10-31
Prevent master controller being disconnected due to backlog

bmerry:
There are two fixes here:
1. Reduce the amount of data burst to the master controller when the product controller shuts down.
2. Increase the maximum permitted backlog.

The first seems like it might be sufficient, but the second is added to ensure we don't have any issues when setting all the gains in parallel.

Closes NGC-1440.

### DIFF
--- a/src/katsdpcontroller/defaults.py
+++ b/src/katsdpcontroller/defaults.py
@@ -64,3 +64,8 @@ KATSDPSIGPROC_TUNE_MATCH = "nearest"
 SHUTDOWN_DELAY = 10.0
 #: Time to wait for rx.device-status sensors to become nominal
 RX_DEVICE_STATUS_TIMEOUT = 30.0
+#: Maximum number of bytes in a connection backlog before disconnecting a
+#: katcp client. This is increased compared to the aiokatcp default because
+#: some sensors are large strings and a batch of sensor updates could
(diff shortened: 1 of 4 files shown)


src/katsdpcontroller/defaults.py
```diff
@@ -64,3 +64,8 @@
 SHUTDOWN_DELAY = 10.0
 #: Time to wait for rx.device-status sensors to become nominal
 RX_DEVICE_STATUS_TIMEOUT = 30.0
+#: Maximum number of bytes in a connection backlog before disconnecting a
+#: katcp client. This is increased compared to the aiokatcp default because
+#: some sensors are large strings and a batch of sensor updates could
+#: temporarily cause a large backlog.
+CONNECTION_MAX_BACKLOG = 256 * 1024 * 1024
```
james-smith-za, line 71: 256 MiB? Seems a lot.

james-smith-za (approved):

src/katsdpcontroller/defaults.py
bmerry: In a 32K, 80A, wide+narrow instrument just the eq sensors could be 400MB (if they're specified with high precision), so if anything this might not be enough.

PR #784, opened 2024-12-04
Add ability to suppress 30s good data timeout

james-smith-za:
Add an option to develop mode to suppress the 30s good data timeout.

The context for this is that one potential reason for a task not receiving good data within 30s of startup is a problem with the network. We probably wouldn't want to run operations under such circumstances, but if we're trying to debug network issues then having the product die after 30s doesn't give enough time.

The FailReply is replaced by an error message in the logger for traceability, but the product should not in fact just die.

Closes NGC-1571

### DIFF
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -237,11 +237,17 @@ class ServiceOverride:
 
 class DevelopOptions:
     def __init__(
-        self, *, any_gpu: bool = False, disable_ibverbs: bool = False, less_resources: bool = False
+        self,
+        *,
+        any_gpu: bool = False,
+        disable_ibverbs: bool = False,
+        less_resources: bool = False,
+        disable_good_data_timeout: bool = False,
     ) -> None:
         self.any_gpu = any_gpu
         self.disable_ibverbs = disable_ibverbs
         self.less_resources = less_resources
+        self.disable_good_data_timeout: bool = False
 
     @classmethod
     def from_config(cls, config: Mapping[str, bool]) -> "DevelopOptions":
@@ -249,11 +255,14 @@ class DevelopOptions:
             any_gpu=config.get("any_gpu", False),
             disable_ibverbs=config.get("disable_ibverbs", False),
             less_resources=config.get("less_resources", False),
+            disable_good_data_timeout=config.get("disable_good_data_timeout", False),
         )
 
     @classmethod
     def from_bool(cls, opt: bool) -> "DevelopOptions":
-        return cls(any_gpu=opt, disable_ibverbs=opt, less_resources=opt)
+        return cls(
+            any_gpu=opt, disable_ibverbs=opt, less_resources=opt, disable_good_data_timeout=opt
+        )
 
 
 class Options:
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -862,17 +862,24 @@ class SubarrayProduct:
             return
 
         logger.info(f"Waiting for {len(sensors)} tasks to receive good data")
+        good_data_received = True
         try:
             await asyncio.wait_for(future, timeout=RX_DEVICE_STATUS_TIMEOUT)
         except asyncio.TimeoutError:
-            raise FailReply(
+            good_data_received = False
+            msg = (
                 f"Some tasks did not receive good data within {RX_DEVICE_STATUS_TIMEOUT}s: "
                 + ", ".join(sorted(missing))
-            ) from None
+            )
+            if self.configuration.options.develop.disable_good_data_timeout:
+                logger.error(msg)
+            else:
+                raise FailReply(msg) from None
         finally:
             for sensor in sensors:
                 sensor.detach(observer)
-        logger.info("All tasks receiving good data")
+        if good_data_received:
+            logger.info("All tasks receiving good data")
 
     async def _capture_done_impl(self, capture_block: CaptureBlock) -> None:
         """Stop a capture block.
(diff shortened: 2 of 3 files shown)


src/katsdpcontroller/product_config.py
```diff
@@ -237,23 +237,32 @@ def from_config(cls, config: Mapping[str, Any]) -> "ServiceOverride":
...
+        *,
+        any_gpu: bool = False,
+        disable_ibverbs: bool = False,
+        less_resources: bool = False,
+        disable_good_data_timeout: bool = False,
     ) -> None:
         self.any_gpu = any_gpu
         self.disable_ibverbs = disable_ibverbs
         self.less_resources = less_resources
+        self.disable_good_data_timeout: bool = False
```
bmerry, line 250: Surely you mean?
```suggestion
        self.disable_good_data_timeout = disable_good_data_timeout
```

src/katsdpcontroller/product_controller.py
```diff
@@ -862,17 +862,24 @@ def observer(sensor, reading):
             return
 
         logger.info(f"Waiting for {len(sensors)} tasks to receive good data")
```
bmerry, line 864: My suggestion is that if the user has asked not to have the timeout, then we shouldn't even try to wait i.e. put the rest of the function from this point inside an `if` statement. If someone is using this option, then there is a high chance it's going to fail, and we might as well not make them wait an extra 30 seconds in their debug cycle.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -586,6 +586,9 @@
                                 "any_gpu": {"type": "boolean", "default": false},
                                 "disable_ibverbs": {"type": "boolean", "default": false},
                                 "less_resources": {"type": "boolean", "default": false}
+{% if version >= "4.2" %}
```
bmerry, line 589: As discussed, add a new version (4.3) for this, and update the changelog in the document.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -586,6 +586,9 @@
...
+                                "disable_good_data_timeout": {"type": "boolean", "default": false}
```
bmerry, line 590: You're missing a comma between the previous item and the next one. You can fix it like this:
```suggestion
                                , "disable_good_data_timeout": {"type": "boolean", "default": false}
```
but I think that looks ugly. I usually try to add new stuff at the top rather than the bottom to avoid this issue.

Something we could consider one day is writing this file in YAML - as long as we stick to the JSON data model and don't use the fancy features of YAML it should make things more readable. It looks like katsdpcontroller sucks in pyyaml indirectly anyway.

bmerry (changes requested):

src/katsdpcontroller/product_config.py
james-smith-za: Oops. Yes.

src/katsdpcontroller/product_controller.py
james-smith-za: I guess the only disadvantage to that is that you don't get the log message but as you point out, if you're using this option, then I guess you already know something like this is coming and it's not necessary. Fair point.

src/katsdpcontroller/schemas/product_config.json.j2
james-smith-za: Makes sense. I'll go with adding at the top for aesthetic reasons.

james-smith-za:
Checks haven't completed yet for the latest commit but they worked for the previous one so I am quite confident that it'll go through.

test/utils.py
```diff
@@ -214,11 +214,15 @@
             }
         }
     },
-    "config": {}
+    "config": {
+        "develop": {
+            "data_timeout": 1.0
```
bmerry, line 219: To be on the safer side I'd rather make this change just in the `test_configure_bad_rx_device_status` test, rather than changing it for all tests.

test/utils.py
```diff
@@ -269,7 +273,11 @@
             "src_pol": 1
         }
     },
-    "config": {}
+    "config": {
+        "develop": {
+            "data_timeout": 1.0
+        }
+    }
```
bmerry, line 280: Do you actually need this change? I'd rather have most tests not specify 'develop', so that tests are more similar to real-world conditions where possible.

bmerry (changes requested):

test/utils.py
bmerry: Your "reversion" has now removed the "config" attribute entirely.

test/test_product_config.py
bmerry, line 1: Please extend the TestOptions tests to test the timeout option.

bmerry (changes requested):

test/test_product_config.py
```diff
@@ -226,6 +227,7 @@ def test_from_config_bool(self) -> None:
         assert options.develop.any_gpu is True
         assert options.develop.disable_ibverbs is True
         assert options.develop.less_resources is True
+        assert options.develop.data_timeout == 30.0
```
bmerry, line 230:
```suggestion
        assert options.develop.data_timeout == defaults.RX_DEVICE_STATUS_TIMEOUT
```
to avoid repeating the magic number.

bmerry (changes requested):
Just one niggle - feel free to merge with that fixed.

test/test_product_config.py
james-smith-za: Fair point.

PR #787, opened 2025-01-21
Add support for product schema version 4.4

bmerry:
This adds `taps` and `window_function` parameters to gpucbf.antenna_channelised_voltage. Pass them through to katgpucbf. The latter requires an update to katgpucbf.

Also change w_cutoff from positive_number to nonneg_number, so that w_cutoff=0 is possible (which effectively disables the sinc in generate_pfb_weights. This is done retroactively for all schema versions, since the specification didn't say what the allowed range was.

Relates to NGC-1004.

### DIFF
--- a/src/katsdpcontroller/schemas/product_config.json.j2
+++ b/src/katsdpcontroller/schemas/product_config.json.j2
@@ -508,7 +508,7 @@
                                     },
                                     "w_cutoff": {"$ref": "#/definitions/nonneg_number", "default": 1.0},
 {% if version >= "4.4" %}
-                                    "window_function": {"$ref": "#/definitions/window_function"},
+                                    "window_function": {"$ref": "#/definitions/window_function", "default": "hann"},
                                     "taps": {"$ref": "#/definitions/positive_integer"},
 {% endif %}
 {% if version >= "3.3" %}
--- a/src/katsdpcontroller/schemas/product_config.json.j2
+++ b/src/katsdpcontroller/schemas/product_config.json.j2
@@ -1,5 +1,5 @@
 {% macro versions() %}
-["3.0", "3.1", "3.2", "3.3", "3.4", "3.5", "4.0", "4.1", "4.2", "4.3"]
+["3.0", "3.1", "3.2", "3.3", "3.4", "3.5", "4.0", "4.1", "4.2", "4.3", "4.4"]
 {% endmacro %}
 
 {% macro validate(version) %}
@@ -90,6 +90,9 @@
         },
         "dither": {
             "enum": ["uniform", "none"]
+        },
+        "window_function": {
+            "enum": ["hann", "rect"]
         }
     },
     "type": "object",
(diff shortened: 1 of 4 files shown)


james-smith-za:
> Also change w_cutoff from positive_number to nonneg_number, so that w_cutoff=0 is possible (which effectively disables the sinc in generate_pfb_weights. This is done retroactively for all schema versions, since the specification didn't say what the allowed range was.

We can probably get away with this since we haven't moved to "production" yet.

src/katsdpcontroller/schemas/product_config.json.j2
```diff
@@ -503,7 +506,11 @@
                                         "items": {"$ref": "#/definitions/stream_name"},
                                         "uniqueItems": true
                                     },
-                                    "w_cutoff": {"$ref": "#/definitions/positive_number", "default": 1.0},
+                                    "w_cutoff": {"$ref": "#/definitions/nonneg_number", "default": 1.0},
+{% if version >= "4.4" %}
+                                    "window_function": {"$ref": "#/definitions/window_function"},
+                                    "taps": {"$ref": "#/definitions/positive_integer"},
```
amishatishpatel, line 512: Admittedly not the most pressing question, just while we're here
* Any particular reason `w_cutoff` has a default but `window_function` and `taps` do not?

I know there are defaults in 
* `katsdpcontroller/product_config.py::GpuCbfAntennaChannelisedVoltage` parameter list, and in
* `fgpu/main.py` for `parse_{wideband, narrowband}`

amishatishpatel (approved):
Seems entirely sane to me.

Re: The latter in
> This adds taps and window_function parameters to gpucbf.antenna_channelised_voltage. Pass them through to katgpucbf. **The latter requires an update to katgpucbf.**

Was that not done in https://github.com/ska-sa/katgpucbf/pull/899 ?

src/katsdpcontroller/schemas/product_config.json.j2
bmerry: Good question. I should probably add the default for window_function. I don't want to add one for taps, because that's implementation-defined rather than part of the schema itself.

It's a fairly arbitrary choice, because the default values defined here have no impact and serve purely as documentation.

amishatishpatel:
> Seems entirely sane to me.
> 
> Re: The latter in
> 
> > This adds taps and window_function parameters to gpucbf.antenna_channelised_voltage. Pass them through to katgpucbf. **The latter requires an update to katgpucbf.**
> 
> Was that not done in [ska-sa/katgpucbf#899](https://github.com/ska-sa/katgpucbf/pull/899) ?

@bmerry also my question above. What other plumbing needs to be done? Perhaps I should just be patient and see what review comes next.

amishatishpatel (approved):

amishatishpatel:
> > > Was that not done in [ska-sa/katgpucbf#899](https://github.com/ska-sa/katgpucbf/pull/899) ?
> > 
> > 
> > @bmerry also my question above. What other plumbing needs to be done? Perhaps I should just be patient and see what review comes next.
> 
> You are correct, Amish, it was done in the PR you referred.
> 
> It's just the loose convention we've been following for traceability to include this reference in the PR if some other change in another repo is required for it to work.

I'm definitely familiar with the convention (very much for it as well). What threw me is that the katgpucbf update happened before this PR. Usually (in my mind, anyway) the literal PR URL or PR number is referenced/included in the blurb.

bmerry:
> I'm definitely familiar with the convention (very much for it as well). What threw me is that the katgpucbf update happened before this PR. Usually (in my mind, anyway) the literal PR URL or PR number is referenced/included in the blurb.

I think it's because I wrote the commit message earlier and raised the PR after testing.

PR #790, opened 2025-02-05
Implement schema version 4.5

bmerry:
This adds the 'vlbi' attribute to 'narrowband', and passes on the
'pass_bandwidth' to katgpucbf.

Also increases the DDC_TAPS_RATIO to 24 to get better filter response.
A functional change is that the window_function will now always be
passed to katgpucbf - in practice that's harder to avoid now that the
default depends on whether 'vlbi' is specified (although it could be
done if necessary).

Closes NGC-1561.

### DIFF
--- a/src/katsdpcontroller/defaults.py
+++ b/src/katsdpcontroller/defaults.py
@@ -48,7 +48,7 @@ GPUCBF_JONES_PER_BATCH = 2**20
 XBGPU_MAX_SRC_DATA_RATE = 5.45e9
 #: Number of polyphase filter-bank taps for gpucbf F-engines
 PFB_TAPS = 16
-#: Multiply by subsamplingfactor to get digitiser down-conversion taps
+#: Multiply by subsampling factor to get digitiser down-conversion taps
 #: for gpucbf F-engines.
 DDC_TAPS_RATIO = 24
 #: Window function for non-VLBI antenna-channelised-voltage products
--- a/src/katsdpcontroller/defaults.py
+++ b/src/katsdpcontroller/defaults.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2013-2023, National Research Foundation (SARAO)
+# Copyright (c) 2013-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
@@ -48,9 +48,13 @@ GPUCBF_JONES_PER_BATCH = 2**20
 XBGPU_MAX_SRC_DATA_RATE = 5.45e9
 #: Number of polyphase filter-bank taps for gpucbf F-engines
 PFB_TAPS = 16
-#: Multiply by decimation factor to get digitiser down-conversion taps
+#: Multiply by subsamplingfactor to get digitiser down-conversion taps
 #: for gpucbf F-engines.
-DDC_TAPS_RATIO = 12
+DDC_TAPS_RATIO = 24
+#: Window function for non-VLBI antenna-channelised-voltage products
+WINDOW_FUNCTION = "hann"
+#: Window function for VLBI antenna-channelised-voltage products
+WINDOW_FUNCTION_VLBI = "rect"
 #: Default payload size for gpucbf data products. Bigger is better to
 #: minimise the number of packets/second to process.
 GPUCBF_PACKET_PAYLOAD_BYTES = 8192
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -500,10 +500,11 @@ def _make_fgpu(
 
         if stream.narrowband is not None:
             ddc_group_delay = -(stream.narrowband.ddc_taps - 1) / 2
-            # Complex-to-complex PFB, but katgpucbf first channelises to
-            # 2N channels then discards N of them.
-            subsampling = stream.narrowband.decimation_factor
-            pfb_group_delay = -(2 * stream.n_chans * stream.pfb_taps - 1) / 2 * subsampling
+            subsampling = stream.narrowband.subsampling
+            internal_channels = stream.n_chans
+            if stream.narrowband.vlbi is not None:
+                internal_channels *= 2
+            pfb_group_delay = -(internal_channels * stream.pfb_taps - 1) / 2 * subsampling
         else:
             ddc_group_delay = 0.0
             # Real-to-complex PFB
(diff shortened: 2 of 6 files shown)


bmerry:
Note that this builds on top of #789 - so ignore the first commit which duplicates that PR.

src/katsdpcontroller/generator.py
```diff
@@ -502,7 +502,7 @@ def _make_fgpu(
             ddc_group_delay = -(stream.narrowband.ddc_taps - 1) / 2
             subsampling = stream.narrowband.subsampling
             internal_channels = stream.n_chans
-            if stream.narrowband.vlbi is not None:
+            if stream.narrowband.vlbi is None:
```
amishatishpatel, line 505: Thank you for fixing this btw.

test/test_product_config.py
```diff
@@ -494,11 +506,44 @@ def test_from_config_narrowband(
...
+        )
+        # Most attributes we don't bother testing - just those affected by VLBI mode
+        assert acv.bandwidth == src_streams[0].adc_sample_rate / 2 / 8
+        assert (
+            acv.n_samples_between_spectra
+            == 2
+            * narrowband_vlbi_config["n_chans"]
+            * narrowband_vlbi_config["narrowband"]["decimation_factor"]
+        )
+        assert acv.data_rate(1.0, 0) == 27392e6 * 2 / 8
```
amishatishpatel, line 530: You're using `/ 8` 
* in the `bandwidth` check,
* the `["decimation_factor"]` (it's `8` in the config fixture), and
* in the `data_rate(...)` check

Even if we ignore the middle example, the first two are ambiguous enough. At least to me, anyway. I thought I was sure what unit `bandwidth` is in but now not so much.

test/test_product_config.py
```diff
@@ -494,11 +506,44 @@ def test_from_config_narrowband(
...
+        assert acv.w_cutoff == 0.0  # Default value for VLBI
+        assert acv.window_function == defaults.WINDOW_FUNCTION_VLBI
+        assert acv.pfb_taps == 1
+
+    def test_vlbi_bad_pass_bandwidth(
+        self,
+        narrowband_vlbi_config: Dict[str, Any],
+        src_streams: List[DigBasebandVoltageStreamBase],
+    ) -> None:
+        bandwidth = src_streams[0].adc_sample_rate / 2 / 8
```
amishatishpatel, line 540: Same `/ 8` comment for here.

amishatishpatel (changes requested):
Apologies for the delay. Just a couple comments on the unit test update.

src/katsdpcontroller/defaults.py
```diff
@@ -48,9 +48,13 @@
 XBGPU_MAX_SRC_DATA_RATE = 5.45e9
 #: Number of polyphase filter-bank taps for gpucbf F-engines
 PFB_TAPS = 16
-#: Multiply by decimation factor to get digitiser down-conversion taps
+#: Multiply by subsamplingfactor to get digitiser down-conversion taps
```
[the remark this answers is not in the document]
bmerry: Oops, fixed.

test/test_product_config.py
bmerry: I've introduced a named constant for this repeated magic value. Arguably I should do it for various other constants (like 1712e6), but maybe this one is more confusing than others.

test/test_product_config.py
```diff
@@ -491,14 +504,42 @@ def test_from_config_narrowband(
...
+        self,
+        narrowband_vlbi_config: Dict[str, Any],
+        src_streams: List[DigBasebandVoltageStreamBase],
+    ) -> None:
+        bandwidth = src_streams[0].adc_sample_rate / 2 / DECIMATION
+        narrowband_vlbi_config["narrowband"]["vlbi"]["pass_bandwidth"] = bandwidth
+        with pytest.raises(ValueError):
+            GpucbfAntennaChannelisedVoltageStream.from_config(
+                Options(), "vlbi1_acv", narrowband_vlbi_config, src_streams, {}
+            )
```
amishatishpatel, line 541: Argh, I'm going to have to be a small pain and request an update to this `pytest.raises` statement.

I was wondering why it was looking so meek. 
* There are some tests where the name doesn't exactly tell me/us what failure case it's checking
  * e.g. `test_n_chans_not_power_of_two` is explicit enough, but 
  * `test_narrowband_bad_centre_frequency` includes the `match={err_string}` to report why ~exactly the arg was bad
* I would request the same treatment here to say why `bandwidth` is bad for `pass_bandwidth`
  * If I need to motivate my request more, I had to go back to `product_config.py` to check why `bandwidth` is bad.
  * I feel like the test should have most/all of the information I need.

amishatishpatel (changes requested):
Sorry, one more small change requested (tweaking the `pytest.raises` statement).

Otherwise, good to go.

test/test_product_config.py
```diff
@@ -528,14 +528,21 @@ def test_from_config_narrowband_vlbi(
...
     ) -> None:
         bandwidth = src_streams[0].adc_sample_rate / 2 / DECIMATION
-        narrowband_vlbi_config["narrowband"]["vlbi"]["pass_bandwidth"] = bandwidth
-        with pytest.raises(ValueError):
+        pass_bandwidth = bandwidth + extra
+        narrowband_vlbi_config["narrowband"]["vlbi"]["pass_bandwidth"] = pass_bandwidth
+        with pytest.raises(
+            ValueError,
+            match=rf"pass_bandwidth \({pass_bandwidth}\) must be strictly less than "
+            rf"bandwidth \({bandwidth}\)",
```
amishatishpatel, line 544: Interesting you use `rf` but also include the backslashes for the normal brackets/braces.

I see Line 1161 (`test_bad_continuum_factor`) also uses `rf"..."` but doesn't include the backslashes for its normal braces. Either approach works? One more sound than the other?

amishatishpatel (approved):
Appreciate the addition of the `extra` in the `bad_pass_bandwidth` test.

test/test_product_config.py
bmerry: The difference is line 1161 wrapped the string in `re.escape` to escape the parentheses, rather than escaping them with backslashes. I like that, so I've made the same change here.

The r prefix means that backslashes are regex escapes instead of Python string escapes. With the backslashes gone it's not needed, so I've removed it.

amishatishpatel: My apologies, you're quite right Re: `re.escape`. Still, appreciate the update/conformity.

PR #792, opened 2025-02-11
Add ?dsim-signals request

bmerry:
This allows a client to issue requests to dsims without needing separate connections to them. This also ensures that when a client receives the reply to ?dsim-signals, the product controller is guaranteed to have received the update of the steady-state-timestamp sensor from the dsim, and hence a subsequent ?capture-start will latch it.

Relates to NGC-1542.


src/katsdpcontroller/product_controller.py
```diff
@@ -2134,6 +2156,26 @@ async def request_beam_delays(self, ctx, stream: str, *coefficient_sets: str) ->
...
+        dsim
+            The task name for the dsim (implementation-defined).
+        signals_str
+            Textual description of the signals. Refer to the katgpucbf manual
+            for details. The description must produce one signal per polarisation.
+
+        Returns
+        -------
+        timestamp
+            First timestamp which will use the new signals
```
amishatishpatel, line 2175: Full stop at the end of this sentence, please. Just to be consistent with your additions above.

src/katsdpcontroller/product_controller.py
```diff
@@ -2134,6 +2156,26 @@ async def request_beam_delays(self, ctx, stream: str, *coefficient_sets: str) ->
...
+    ) -> int:
+        """Update the signals that are generated by a dsim.
+
+        Parameters
+        ----------
+        dsim
+            The task name for the dsim (implementation-defined).
+        signals_str
+            Textual description of the signals. Refer to the katgpucbf manual
+            for details. The description must produce one signal per polarisation.
```
amishatishpatel, line 2170: Is it worth adding a description for `period` here? I genuinely don't know. You do refer to the katgpucbf manual for `signals_str`; could you perhaps somehow include a note for `period` in there as well?

src/katsdpcontroller/product_controller.py
```diff
@@ -1497,6 +1497,28 @@ async def beam_delays(self, stream_name: str, coefficient_sets: Sequence[str]) -
...
+        for task in self.physical_graph:
+            if (
+                isinstance(task, tasks.ProductPhysicalTaskMixin)
+                and task.logical_node.task_type == "sim"
+                and task.name == dsim
+            ):
+                dsim_task = task
+                break
+        else:
+            raise FailReply(f"No dsim named {dsim}")
```
amishatishpatel, line 1512: Trying to compare notes and files my side, can't quite get to the bottom of my own question.
* Any particular reason you can't do the usual `stream = self._find_stream(...)`?
* I gather it's something along the lines of not needing to issue a `_multi_request` because it's direct to the dsim?

amishatishpatel (approved):
No major observations/complaints. Just one question for my own understanding.

src/katsdpcontroller/product_controller.py
bmerry: Good point. I'll fix that.

src/katsdpcontroller/product_controller.py
bmerry: This request is a bit awkward because there isn't a one-to-one mapping between streams and dsims. generator.py goes and finds two streams (e.g. m000h and m000v) which have the same antenna and sample rate, and runs them within one dsim.

amishatishpatel (approved):

PR #794, opened 2025-02-18
Add ?dsim-time request

bmerry:
This relates to NGC-1542: it allows CBF qualification tests to query the dsim for time without having a direct katcp connection.

### DIFF
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -1500,6 +1500,11 @@ class SubarrayProduct:
     def _get_dsim_katcp(self, dsim: str) -> aiokatcp.Client:
         """Get the katcp client for connecting to a dsim.
 
+        Parameters
+        ----------
+        dsim
+            The task name for the dsim
+
         Raises
         ------
         FailReply
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -1497,7 +1497,15 @@ class SubarrayProduct:
             timeout=DELAYS_TIMEOUT,
         )
 
-    async def dsim_signals(self, dsim: str, signals_str: str, period: Optional[int] = None) -> int:
+    def _get_dsim_katcp(self, dsim: str) -> aiokatcp.Client:
+        """Get the katcp client for connecting to a dsim.
+
+        Raises
+        ------
+        FailReply
+            if the dsim name is not known, there is no connection to it, or the
+            subarray product is not in a suitable state.
+        """
         if self.state not in {ProductState.CAPTURING, ProductState.IDLE}:
             raise FailReply(f"Cannot set dsim signals in state {self.state}")
         for task in self.physical_graph:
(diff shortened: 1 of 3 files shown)


src/katsdpcontroller/product_controller.py
```diff
@@ -1497,7 +1497,15 @@ async def beam_delays(self, stream_name: str, coefficient_sets: Sequence[str]) -
...
-    async def dsim_signals(self, dsim: str, signals_str: str, period: Optional[int] = None) -> int:
+    def _get_dsim_katcp(self, dsim: str) -> aiokatcp.Client:
+        """Get the katcp client for connecting to a dsim.
+
+        Raises
+        ------
+        FailReply
+            if the dsim name is not known, there is no connection to it, or the
+            subarray product is not in a suitable state.
+        """
```
amishatishpatel, line 1508: What's the protocol for including a parameter description for a param that will _always_ come from a public-facing, documented call?
* Basically, I previously conceded we didn't need to document `dsim: str` param in `dsim_signals` because it came from `request_dsim_signals`

Would you be opposed to adding that same description here?
```
Parameters
------------
dsim
    The task name for the dsim (implementation-defined).
```

src/katsdpcontroller/product_controller.py
```diff
@@ -1497,7 +1497,15 @@ async def beam_delays(self, stream_name: str, coefficient_sets: Sequence[str]) -
...
         )
 
-    async def dsim_signals(self, dsim: str, signals_str: str, period: Optional[int] = None) -> int:
+    def _get_dsim_katcp(self, dsim: str) -> aiokatcp.Client:
+        """Get the katcp client for connecting to a dsim.
+
+        Raises
+        ------
+        FailReply
+            if the dsim name is not known, there is no connection to it, or the
```
amishatishpatel, line 1506: Also, capital 'I' for first word "if".

src/katsdpcontroller/product_controller.py
```diff
@@ -1510,14 +1518,23 @@ async def dsim_signals(self, dsim: str, signals_str: str, period: Optional[int]
                 break
         else:
             raise FailReply(f"No dsim named {dsim}")
+        if dsim_task.katcp_connection is None:
+            raise FailReply(f"No katcp connection to {dsim}")
```
amishatishpatel, line 1522: I've always meant to check
* How does this scenario transpire?
* Why don't we induce/create our own connection?

test/test_product_controller.py
```diff
@@ -832,6 +833,18 @@ async def test_dsim_signals_bad_dsim(self, client: aiokatcp.Client) -> None:
             await client.request("dsim-signals", "spam", "1;2;")
         await client.request("product-deconfigure")
 
+    async def test_dsim_time(self, client: aiokatcp.Client) -> None:
+        await client.request("product-configure", SUBARRAY_PRODUCT, CONFIG)
+        reply, _ = await client.request("dsim-time", "sim.m900.1712000000.0")
+        assert reply == [b"123456789.5"]
```
amishatishpatel, line 839: Oh this is very neat, with the mocking of `time.time`.

amishatishpatel (approved):
One explicit change (to capitalise a letter). Another question or two.

src/katsdpcontroller/product_controller.py
bmerry: I prefer to use lowercase here, because this will render (or would, or katsdpcontroller actually built docs) as
"FailReply — if the dsim name is not known, ..." So the "if" is not the start of the sentence. It currently looks like katsdpcontroller doesn't have any consistency in this though.

src/katsdpcontroller/product_controller.py
bmerry: If the dsim dies we'll set the `katcp_connection` to None.

bmerry:
Let me know what you think about your suggested changes - I've explained why I did things my way, but I don't feel that strongly so if you'd still like them changed I'm happy to do so.

src/katsdpcontroller/product_controller.py
amishatishpatel: Ah, noted. I'll concede to leave it as lowercase.

src/katsdpcontroller/product_controller.py
amishatishpatel: I'm going to vote for the docstring update to include the parameter description. Just for anyone else/implementing the next `dsim-request`, you don't need to consult existing call chains on what exactly it's expecting.

PR #795, opened 2025-02-19
Add more sensors to describe fgpu configuration

bmerry:
See NGC-1605.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -539,7 +539,8 @@ def _make_fgpu(
             Sensor(
                 float,
                 f"{stream.name}.pass-bandwidth",
-                "Bandwidth over which the CBF produces a flat response" "Hz",
+                "Bandwidth over which the CBF produces a flat response",
+                "Hz",
                 default=stream.pass_bandwidth,
                 initial_status=Sensor.Status.NOMINAL,
             ),
@@ -990,7 +991,8 @@ def _make_xbgpu(
             Sensor(
                 float,
                 f"{stream.name}.pass-bandwidth",
-                "Bandwidth over which the CBF produces a flat response" "Hz",
+                "Bandwidth over which the CBF produces a flat response",
+                "Hz",
                 default=acv.pass_bandwidth,
                 initial_status=Sensor.Status.NOMINAL,
             ),
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -537,6 +537,13 @@ def _make_fgpu(
                 initial_status=Sensor.Status.NOMINAL,
             ),
             Sensor(
+                float,
+                f"{stream.name}.pass-bandwidth",
+                "Bandwidth over which the CBF produces a flat response" "Hz",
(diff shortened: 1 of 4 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -536,6 +536,13 @@ def _make_fgpu(
                 default=stream.bandwidth,
                 initial_status=Sensor.Status.NOMINAL,
             ),
+            Sensor(
+                float,
+                f"{stream.name}.pass-bandwidth",
+                "Bandwidth over which the CBF produces a flat response" "Hz",
```
amishatishpatel, line 542: Not sure what happened here. This needs to be two separate args
```
Sensor(
    ...,
    "Bandwidth over which the CBF [...]",
    "Hz",
    ...,
)
```

src/katsdpcontroller/generator.py
```diff
@@ -942,6 +987,13 @@ def _make_xbgpu(
                 default=acv.bandwidth,
                 initial_status=Sensor.Status.NOMINAL,
             ),
+            Sensor(
+                float,
+                f"{stream.name}.pass-bandwidth",
+                "Bandwidth over which the CBF produces a flat response" "Hz",
```
amishatishpatel, line 993: Same comment here Re: Making `"Hz"` the unit arg for `Sensor(...)`.

src/katsdpcontroller/product_config.py
```diff
@@ -671,6 +673,11 @@ def subsampling(self) -> int:
         return self.decimation_factor if self.vlbi is None else 2 * self.decimation_factor
 
 
+class GpucbfWindowFunction(enum.Enum):
+    HANN = "hann"
+    RECT = "rect"
```
amishatishpatel, line 678: I feel a bit iffy that 
* you've introduced an enum with a string value here, but
* [fgpu uses integer values in its version](https://github.com/ska-sa/katgpucbf/blob/main/src/katgpucbf/fgpu/output.py#L29-L34)

I mean, not the end of the world seeing as the members have the same name, right?

amishatishpatel (changes requested):
Minor updates to some sensor declarations, please.

src/katsdpcontroller/generator.py
bmerry: Oops: I left out a comma, and Black then wrapped the two separate lines into one. I tend to believe that Python allowing string concatenation like this was a mistake, because it leads to bugs like this. Fixed now.

amishatishpatel: >  I tend to believe that Python allowing string concatenation like this was a mistake, because it leads to bugs like this.

Right? I didn't know it could be done this why. Tried it out locally and it worked, but figured you wouldn't do something like that.

src/katsdpcontroller/product_config.py
amishatishpatel: Noted all around. I am definitely in support of at least updating fgpu to use the same values as here. Agreed on the usefulness of `parse_enum`.

amishatishpatel (approved):

PR #796, opened 2025-02-24
Fix typo in weight-pass sensor description

bmerry:


bmerry:
@amishatishpatel gave an over-the-shoulder review, so merging.

PR #797, opened 2025-02-25
Revive CBF-SDP integration testing

bmerry:
Allow SDP and CBF to run in a single subarray product while using real antennas (or antennas produced by external dsims). The main challenge is that ingest requires mXXX_observer telstate entries to determine the lengths of baselines, and the antenna positions would not normally be available in such a configuration. Allow it to be smuggled in by providing antenna descriptions instead of just names in dig.baseband_voltage streams.

Relates to NGC-1551.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -437,6 +437,15 @@ def _make_dsim(
     return dsim
 
 
+def _make_dig(g: networkx.MultiDiGraph, stream: product_config.DigBasebandVoltageStream) -> None:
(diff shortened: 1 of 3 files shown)


bmerry:
Marking this as a draft because it will still need some unit tests.

ludwigschwardt (approved):
I like the cunning workaround!

bmerry:
@ludwigschwardt could you cast an eye over the unit test I added?

ludwigschwardt (approved):
Looks good.

PR #798, opened 2025-03-10
In --interface-mode, bind PC to same address as MC

bmerry:
This allows the product controllers to be reached externally when the master controller is bound to a non-localhost address.

When no bind address is specified, product controllers will now bind to the wildcard address, but the reported address will be localhost.

Closes NGC-1640.

### DIFF
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2013-2023, National Research Foundation (SARAO)
+# Copyright (c) 2013-2023, 2025, National Research Foundation (SARAO)
 #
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/master_controller.py
```diff
@@ -1967,7 +1967,7 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
     usage = "%(prog)s [options] zk singularity"
     parser = argparse.ArgumentParser(usage=usage)
     parser.add_argument(
-        "-a", "--host", default="", metavar="HOST", help="attach to server HOST [localhost]"
+        "-a", "--host", default="", metavar="HOST", help="bind to server HOST [any]"
```
bmerry, line 1970: This comment was wrong from the start - not related to the functionality change in this PR.

amishatishpatel (approved):

PR #803, opened 2025-04-03
Do DNS resolution on Mesos offers

bmerry:
Resolve hostnames to IP addresses when offers are received, and do not use an offer until we have the IP address. This address can then be used in various other places that were doing their own DNS resolution.

This eliminates a number of failure modes. In particular, the endpoint sensors are now guaranteed to be nominal.

Closes NGC-1531.

### DIFF
--- a/src/katsdpcontroller/scheduler.py
+++ b/src/katsdpcontroller/scheduler.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2013-2024, National Research Foundation (SARAO)
+# Copyright (c) 2013-2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
@@ -337,7 +337,7 @@ class OrderedEnum(Enum):
         return NotImplemented
 
 
-async def poll_ports(host, ports):
+async def poll_ports(address, ports):
     """Waits until a set of TCP ports are accepting connections on a host.
 
     It repeatedly tries to connect to each port until a connection is
@@ -347,8 +347,8 @@ async def poll_ports(host, ports):
 
     Parameters
     ----------
-    host : str
-        Hostname or IP address
+    address : str
+        IP address
     ports : list
         Port numbers to connect to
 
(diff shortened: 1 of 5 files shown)


bmerry:
@amishatishpatel I'd suggest we look at this together next week so I can go over the core of the change with you. It has some subtleties and I think it would help me to explain it to someone to be sure I've covered all the edge cases.

ludwigschwardt:
Excellent idea.

src/katsdpcontroller/scheduler.py
```diff
@@ -3778,6 +3790,92 @@ def set_driver(self, driver):
...
+            pass
+        except Exception:
+            logger.warning(
+                "Failed to resolve %s for offer %s",
+                offer.hostname,
+                offer.id.value,
+                exc_info=True,
+            )
+        else:
+            # If we get there, the resolution succeeded, and the offer is no
```
bmerry, line 3871:
```suggestion
            # If we get here, the resolution succeeded, and the offer is no
```

src/katsdpcontroller/scheduler.py
```diff
@@ -3778,6 +3790,92 @@ def set_driver(self, driver):
...
+            # longer our responsibility.
```
bmerry, line 3872: Or we hit the else path in _resolve_offer, in which case _resolve_offer already declined and we don't need to do it again.

src/katsdpcontroller/scheduler.py
```diff
@@ -3778,6 +3790,92 @@ def set_driver(self, driver):
...
+                        "Failure to resolve address for %s (%s). Waiting 1s to retry.",
+                        hostname,
+                        error,
+                    )
+                    await asyncio.sleep(1)
+                else:
+                    raise
+
+    async def _resolve_offer(self, offer):
+        """Resolve the hostname of an offer to an IP address and record the offer."""
```
bmerry, line 3822: Note that the input offer is modified in place.

test/test_scheduler.py
```diff
@@ -2101,6 +2099,57 @@ async def test_offer_rescinded(self, fix: "TestScheduler.Fixture") -> None:
...
+        assert fix.sched._resolving_offers == {}  # Must clean up the tasks
+        assert fix.nodes[0].state == TaskState.STARTING  # Must not start
+        assert await fix.driver_calls() == [
+            mock.call.declineOffer([offers[0].id]),
+            mock.call.declineOffer([offers[1].id]),
+        ]
+        launch.cancel()
+
+    async def test_decline_unresolved_offers(self, fix: "TestScheduler.Fixture", mocker) -> None:
+        """Test that when an unresolved offer is no longer needed, it is declined and removed."""
```
amishatishpatel, line 2136: If I'm really being nitpicky: There is some minor ambiguity in this docstring.
* Test name outlines `unresolved_offers` plural, which is what is actually tested.
* I guess the docstring/description refers to the general process of what happens to an unresolved offer.

I mainly mention this because I got the impression that one offer would resolve and the other would not.
* Might be the wrong impression, so please do correct me accordingly.

src/katsdpcontroller/tasks.py
```diff
@@ -622,15 +625,7 @@ async def resolve(self, resolver, graph, image=None):
...
-            try:
-                addrinfo = await asyncio.get_event_loop().getaddrinfo(self.host, value)
-                host, port = addrinfo[0][4][:2]
-                endpoint_sensor.set_value(aiokatcp.Address(ipaddress.ip_address(host), port))
-            except socket.gaierror as error:
-                self.logger.warning("Could not resolve %s: %s", self.host, error)
-                endpoint_sensor.set_value(
-                    aiokatcp.Address(ipaddress.IPv4Address("0.0.0.0")), status=Sensor.Status.FAILURE
-                )
+            endpoint_sensor.set_value(aiokatcp.Address(ipaddress.ip_address(self.address), value))
```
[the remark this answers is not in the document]
bmerry: Yes. resolve() is only called after the task has been allocated resources, which means that we know which Agent it is going to run on, and the Agent instance is only built from resolved offers.

test/test_scheduler.py
```diff
@@ -2101,6 +2099,57 @@ async def test_offer_rescinded(self, fix: "TestScheduler.Fixture") -> None:
...
+
+    async def test_offer_resolve_fail(self, fix: "TestScheduler.Fixture", mocker) -> None:
+        """Test receiving an offer that cannot be resolved with getaddrinfo."""
+        mocker.patch.object(asyncio.get_running_loop(), "getaddrinfo", side_effect=getaddrinfo_fail)
+        launch, kill = await fix.transition_node0(TaskState.STARTING)
+        offers = fix.make_offers()
+        fix.sched.resourceOffers(fix.driver, offers)
+        await asyncio.sleep(6)  # Give it a chance to try multiple times
+        assert fix.sched._resolving_offers == {}  # Must clean up the tasks
+        assert fix.nodes[0].state == TaskState.STARTING  # Must not start
```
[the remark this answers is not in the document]
bmerry: I think there was a little copy-and-paste in these, but it just gives a bit more context for why we're asserting this e.g. in the first case it clarifies that we expect `_resolving_offers` to get populated then cleaned up, rather than that it should never get populated.

test/test_scheduler.py
bmerry: Changed the docstring to plural.

amishatishpatel (approved):
Nice one.

PR #804, opened 2025-04-03
Allow mirroring of sensors to the master controller be disabled

bmerry:
Changes to the product-configure schema are proposed edits in https://docs.google.com/document/d/15ZIyuBf4Vk4ESIEdKsQxdSBoLEwh85eyKeDJouLRAZY/edit?pli=1&tab=t.0#heading=h.v512w2fvji5f.

A side effect in the original case is that now product-configure will typically issue three `#interface-changed` mass informs instead of two:
1. Immediately in response to the request, which notified consumers that the `<product>.host` and `<product>.katcp-address` sensors have been added (with the address of the product controller).
2. After connecting to the product controller, and querying the sensors that it has so far.
3. After the product is fully configured.

The first is new, while the previous two already existed. When mirror_sensors is false, only the first one will be issued, because no sensors are proxied from the subarray product. I think this reasonable to add, since otherwise the `.host` and `.katcp-address` sensors might not get recorded if there was an issue connecting to the subarray product.

Similarly, there may be an additional `#interface-changed` mass inform when tearing down the array.

### DIFF
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -437,6 +437,7 @@ class ProductManagerBase(Generic[_P]):
     def _update_products_sensor(self) -> None:
         self._server.sensors["products"].value = json.dumps(sorted(self._products.keys()))
         self._update_device_status()
+        self._server.mass_inform("interface-changed", "sensor-list")
 
     def _add_product(self, product: _P) -> None:
         """Used by subclasses to add a newly-created product."""
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -185,11 +185,16 @@ class Product:
         hostname: str,
         host: _IPAddress,
         ports: Dict[str, int],
-    ) -> None:
+    ) -> bool:
         """Notify product of the location of the katcp interface.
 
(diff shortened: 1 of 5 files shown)


bmerry:
@rishadebrahim I've added you to the PR to review the change in the interface and in the observable functionality, rather than for code review.

amishatishpatel (commented):
Seems sound enough. I guess I'm just not entirely/at all familiar with consumers of `#interface-changed sensor-list` mass informs (and what they do with such info).
* Specifically, Re: Notifying "consumers that the `<product>.host` and `<product>.katcp-address` sensors have been added".

I think your reasoning for the additional `#interface-changed sensor-list` is sound because the status of the master controller could depend on the constituent product.

Before I explicitly Approve, I gather you've run the unit test and confirmed all is well?
* I mainly ask because sdp-jenkins hasn't built this PR.

amishatishpatel (approved):
> the product controller avoids sending an #interface-changed inform every time one of the engines comes online with its own sensors, and just sends one once the whole product is up.

Noted, and thank you for clarifying. I do recall some discussion around/near this during a debugging session a few months ago (client timeout due to sensor updates taking too long?).

PR #806, opened 2025-04-16
Fix TimeoutError

bmerry:
Fixes NGC-1678. There are two parts to this:

1. Fix a race condition that led to a cryptic TimeoutError under some circumstances with insufficient resources, instead of a more useful diagnosis of why the resources were insufficient.
2. Perform DNS resolution on a batch of offers as a batch and only pass the results to the main offer pool once the whole batch has been resolved (or has failed to resolve). To make this robust and avoid holding up good DNS results if one offer is failing, add a timeout to DNS resolution, and reduce the time between attempts. This avoids repeatedly trying to launch on a subset of offers as DNS results dribble in, which is both slow and risks suboptimal placement.

The batch of offers is split up by role, because we may need to cancel offers for a specific role but don't want to cancel the rest of the batch at the same time.

### DIFF
--- a/src/katsdpcontroller/scheduler.py
+++ b/src/katsdpcontroller/scheduler.py
@@ -3996,7 +3996,12 @@ class Scheduler(SchedulerBase, pymesos.Scheduler):
         if offer_id.value in self._resolving_offers:
             resolving_offer = self._resolving_offers.pop(offer_id.value)
             resolving_offer.rescinded.add(offer_id.value)
-            resolving_offer.task.cancel()
+            # Note: we can't cancel the whole _resolving_offer task, because
+            # it may involve other offers that haven't been rescinded. We
+            # could track the individual tasks doing the per-offer resolution,
+            # but that would add complexity, and those tasks have bounded
+            # lifetimes anyway due to timeouts, so it's not critical to
+            # cancel them.
             return  # No point checking self._offers if we haven't resolved it yet
         # TODO: this is not very efficient. A secondary lookup from offer id to
         # the relevant offer info would speed it up.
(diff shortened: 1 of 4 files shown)


src/katsdpcontroller/scheduler.py
```diff
@@ -3951,7 +3995,7 @@ def inverseOffers(self, driver, offers):
     def offerRescinded(self, driver, offer_id):
         if offer_id.value in self._resolving_offers:
             resolving_offer = self._resolving_offers.pop(offer_id.value)
-            resolving_offer.rescinded = True
+            resolving_offer.rescinded.add(offer_id.value)
             resolving_offer.task.cancel()
```
amishatishpatel, line 3999: @bmerry this shouldn't be here anymore - will cancel resolution on *all* lookups.

src/katsdpcontroller/scheduler.py
```diff
@@ -3782,104 +3807,120 @@ def set_driver(self, driver):
...
                     raise
 
-    async def _resolve_offer(self, offer):
-        """Resolve the hostname of an offer to an IP address and record the offer.
+    async def _resolve_offers(self, offers):
+        """Resolve the hostnames of offers to IP address and record the offers.
 
-        Note that the `offer` parameter is modified *in place* to create an
-        `address` element.
+        Note that each offers is modified *in place* to create an `address`
```
amishatishpatel, line 3844: "Note that each *offer in `offers`..." ?

amishatishpatel (approved):
Must admit, reaching my limits of understanding with the updated logic - especially with the new unit tests.

Overall, following the high-level update. Just one docstring tweak for you to consider. Otherwise, good to go.

PR #807, opened 2025-05-07
Handle consul timeouts better

bmerry:
- Catch, log and ignore the exception (so that the subarray product
  continues), as for connection errors.
- Increase the timeout from 5 to 15s to make timing out less likely even
  if the consul master is undergoing a backup.

Closes NGC-1674.

### DIFF
--- a/src/katsdpcontroller/consul.py
+++ b/src/katsdpcontroller/consul.py
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (c) 2013-2023, National Research Foundation (SARAO)
+# Copyright (c) 2013-2023, 2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
@@ -16,6 +16,7 @@
 
 """Dynamically register and deregister services with consul."""
 
+import asyncio
 import logging
 import uuid
 from typing import Optional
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/consul.py
```diff
@@ -27,6 +28,10 @@
 
 CONSUL_PORT = 8500
 CONSUL_URL = yarl.URL.build(scheme="http", host=LOCALHOST, port=CONSUL_PORT)
+# Usually needs must less than this, but registering a service requires
```
amishatishpatel, line 31: *much less than this ?

src/katsdpcontroller/consul.py
```diff
@@ -27,6 +28,10 @@
...
+# storing the record on disk in the consul catalogue, and things like VM
+# backups can make that unusually slow.
+CONSUL_TIMEOUT = 20
```
amishatishpatel, line 34: To confirm, your PR blurb says
> Increase the timeout from 5 to 15s

Just want to make sure this value is correct/as intended.

amishatishpatel (changes requested):
Seems straightforward enough. Just one typo to correct, please.

src/katsdpcontroller/consul.py
bmerry: Ah, I was dithering over whether to make it 15s or 20s. I'll change it to 15s: now that you've cut down the qualification testing time, we shouldn't encounter the backup stall any more, so we don't need to be too conservative.

amishatishpatel (approved):

PR #808, opened 2025-06-02
Use aiokatcp sensor filtering to implement mirror_sensors

bmerry:
This makes mirror_sensors less total: we still get to mirror product-level sensors like device-status and gui-urls. The former lets us update the master-level device-status, and the latter is needed to produce the list of GUIs on the / of the HTTP server.

Relates to NGC-1660.

### DIFF
--- a/test/test_master_controller.py
+++ b/test/test_master_controller.py
@@ -841,14 +841,16 @@ class TestDeviceServer:
         await client.request("product-configure", "product", config_str)
         await exhaust_callbacks()
         interface_changed_callback.assert_called_with(b"sensor-list")
-        if mirror_sensors:
-            expected_product_sensors = (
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/sensor_proxy.py
```diff
@@ -69,6 +70,13 @@ class CloseAction(enum.Enum):
     UNREACHABLE = 2  #: Change the sensor statuses to unreachable
 
 
+class _FilterPredicate(Protocol):
```
amishatishpatel, line 73: Good word choice here.

amishatishpatel (approved):
Can't say I've spotted anything concerning or otherwise questionable.

Worth asking if this requires a unit test update anywhere?

amishatishpatel (approved):

PR #810, opened 2025-07-09
Rename sensor state -> product-state

bmerry:
See NGC-1713

### DIFF
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -517,7 +517,7 @@ class SubarrayProduct:
         )
         self._state_sensor = Sensor(
             ProductState,
-            "state",
+            "product-state",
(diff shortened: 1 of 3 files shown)


src/katsdpcontroller/product_controller.py
```diff
@@ -517,7 +517,7 @@ def __init__(
         )
         self._state_sensor = Sensor(
             ProductState,
-            "state",
+            "product-state",
```
ludwigschwardt, line 520: The internal object is conveniently also called `ProductState`, which confirms that the new name is a good one :-)

bmerry:
@KimMcAlpine since the test was successful, can we merge this?

KimMcAlpine:
> @KimMcAlpine since the test was successful, can we merge this?

Sure, merge away

PR #812, opened 2025-07-30
Add dither-seed sensors for MK+ CBF

amishatishpatel:
This involved:
* Adding sensor-renames to generator.py, and
* Updating the corresponding fake servers.

The logic may seem odd in `generator.py::_make_fgpu`, but I decided to use the 
existing for-loops rather than re-spin the nesting for the `dither-seed` rename.

Contributes to: NGC-1418.

### DIFF
--- a/src/katsdpcontroller/fake_servers.py
+++ b/src/katsdpcontroller/fake_servers.py
@@ -164,6 +164,8 @@ class FakeFgpuDeviceServer(FakeDeviceServer):
                     str,
                     f"{output}.dither-seed",
                     "Random seed used in dithering for quantisation",
+                    # This is the largest value that katgpucbf can return,
+                    # which is also the most likely to break something.
                     default=f"{2**64 - 1}",
                     initial_status=Sensor.Status.NOMINAL,
                 )
@@ -339,6 +341,8 @@ class FakeXbgpuDeviceServer(FakeDeviceServer):
                     str,
                     f"{beam_name}.dither-seed",
                     "Random seed used in dithering for quantisation",
+                    # This is the largest value that katgpucbf can return,
+                    # which is also the most likely to break something.
                     default=f"{2**64 - 1}",
                     initial_status=Sensor.Status.NOMINAL,
                 )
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/fake_servers.py
```diff
@@ -157,8 +157,18 @@ def __init__(self, *args, **kwargs) -> None:
...
-        for pol in range(self.N_POLS):
-            for output in self._output_names:
+
+        for output in self._output_names:
+            self.sensors.add(
+                Sensor(
+                    str,
+                    f"{output}.dither-seed",
+                    "Random seed used in dithering for quantisation",
+                    default=f"{2**64 - 1}",
```
bmerry, line 167: Maybe add a comment that you're using the largest value that katgpucbf could return here, since that's the most likely value to break anything.

src/katsdpcontroller/generator.py
```diff
@@ -916,6 +916,15 @@ def _make_fgpu(
...
                         f"{stream.name}.input{j}.{name}"
                     ] = f"{stream.name}.{label}.{name}"
+                for name in [
+                    "dither-seed",
+                ]:
+                    # These engine-level sensors get replicated as well, but are not per-input
+                    orig_sensor_name = f"{stream.name}.{name}"
+                    renames = fgpu.sensor_renames.get(orig_sensor_name, [])
+                    renames.append(f"{stream.name}.{label}.{name}")
+                    fgpu.sensor_renames[orig_sensor_name] = renames
```
bmerry, line 926: As noted in the general discussion I'd rather just add the right value to sensor_renames once rather than iteratively appending. But as a general tip, you can do this more simply like this (untested so I may have the syntax slightly wrong):
```suggestion
                    # These engine-level sensors get replicated as well, but are not per-input
                    fgpu.sensor_renames[f"{stream.name}.{name}"].setdefault([]).append(f"{stream.name}.{label}.{name}")
```

amishatishpatel: Fair enough. I always seem to go the wrong way on these 50/50s.

PR #833, opened 2026-03-25
Add static Prometheus gauges for vgpu_expected_*

bmerry:
This will allow the expected rate to be plotted on a Grafana dashboard (NGC-1924).

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1787,6 +1787,11 @@ def _make_vgpu(
                     f"{pol}{channel}.{perthread_name}"
                 ] = f"{stream.name}.{pol}{channel}.{perthread_name}"
 
+    vgpu.static_gauges["vgpu_expected_input_heaps_per_second"] = sum(
+        s.n_substreams / s.int_time for s in tacv
+    )
+    vgpu.static_gauges["vgpu_expected_engines"] = 1.0
+
     _add_task_sensors(g, [stream], [vgpu.name])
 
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1787,6 +1787,11 @@ def _make_vgpu(
                     f"{pol}{channel}.{perthread_name}"
                 ] = f"{stream.name}.{pol}{channel}.{perthread_name}"
 
+    vgpu.static_gauges["vgpu_expected_input_heaps_per_second"] = sum(
+        s.n_substreams / s.int_time for s in tacv
```
amishatishpatel, line 1791: Sorry I know I'm being a bit lazy here: Can you explain why this calculation is a "substreams per int-time"?

I think I may have gotten a helping hand in understanding from vscode's Copilot extension. 
- It reminded me that `s.int_time` is in the expected/required `self.spectra_per_heap * self.n_samples_between_spectra / self.adc_sample_rate` (whose units are altogether more familiar)
- So this value is "across all substreams" from the TACV stream?

amishatishpatel (commented):
One comment for ~clarification before I explicitly Approve.

src/katsdpcontroller/generator.py
bmerry: int_time is the time between batches, so 1/int_time is the rate of batches. And each batch contains n_substreams heaps.

amishatishpatel (approved):

PR #834, opened 2026-05-06
Make ?product-configure log exceptions

amishatishpatel:
This was specifically intended to capture the "Failed to configure within _timeout_ ",
but it will catch anything within `product_configure`.

I manually built an image with a reduced timeout of 10s to obtain the screenshot below.
- http://cbf-mc.cbf.mkat.karoo.kat.ac.za:5601/app/logtrail#/?q=*&h=All&t=15:30:20&i=logstash-*&_g=()
image

Apologies for the untidy commit history...
- I realised I didn't need to explicitly pass the exception to the logger.error call

Resolves: NGC-1892.

### DIFF
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -1608,7 +1608,7 @@ class DeviceServer(aiokatcp.DeviceServer):
         except Exception as exc:
             logger.error(
                 f"Failed to configure product {name}: {exc}",
-                exc_info=exc,
+                exc_info=True,
                 extra=dict(subarray_product_id=name),
             )
             raise FailReply(f"Failed to configure product {name}: {exc}") from exc
--- a/src/katsdpcontroller/master_controller.py
+++ b/src/katsdpcontroller/master_controller.py
@@ -1608,7 +1608,7 @@ class DeviceServer(aiokatcp.DeviceServer):
         except Exception as exc:
             logger.error(
                 f"Failed to configure product {name}: {exc}",
-                exc_info=True,
+                exc_info=exc,
                 extra=dict(subarray_product_id=name),
             )
             raise FailReply(f"Failed to configure product {name}: {exc}") from exc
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/master_controller.py
```diff
@@ -1603,7 +1603,15 @@ async def request_product_configure(self, ctx, name: str, config: str) -> Tuple[
             config_dict = load_json_dict(config)
         except ValueError as exc:
             raise FailReply(f"Config is not valid JSON: {exc}") from None
-        product = await self.product_configure(name, config_dict)
+        try:
+            product = await self.product_configure(name, config_dict)
+        except Exception as exc:
+            logger.error(
+                f"Failed to configure product {name}: {exc}",
```
bmerry:
```suggestion
                "Failed to configure product %s: %s",
                name,
                exc,
```

src/katsdpcontroller/master_controller.py
```diff
@@ -1603,7 +1603,15 @@ async def request_product_configure(self, ctx, name: str, config: str) -> Tuple[
             config_dict = load_json_dict(config)
         except ValueError as exc:
             raise FailReply(f"Config is not valid JSON: {exc}") from None
-        product = await self.product_configure(name, config_dict)
+        try:
+            product = await self.product_configure(name, config_dict)
+        except Exception as exc:
+            logger.error(
```
bmerry: You can use `logger.exception` to get the exception traceback without needing to specify exc_info=True.

src/katsdpcontroller/master_controller.py
```diff
@@ -1603,7 +1603,15 @@ async def request_product_configure(self, ctx, name: str, config: str) -> Tuple[
...
+                f"Failed to configure product {name}: {exc}",
+                exc_info=True,
+                extra=dict(subarray_product_id=name),
+            )
+            raise FailReply(f"Failed to configure product {name}: {exc}") from exc
```
bmerry: I don't think we should be modifying the exception message here: one of the possible exception messages is already `f"Failed to configure {name}: {exc}"`. So here just

```suggestion
            raise
```

The FailReply is tied to a request, so the caller should already know what subarray product it is, and if we find it gets lost somewhere, then we can add the information at the point the original error message is formed.

bmerry (changes requested):

bmerry (approved):

PR #835, opened 2026-05-18
NGC-1975 bump packages versions

vincentseotlo:
Upgrades only the packages that does not require python upgrade

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -90,7 +90,7 @@ docker==7.0.0
     # via katsdpcontroller (setup.cfg)
 ephem==4.1.5
     # via katpoint
-flask==3.0.2
+flask==3.0.3
     # via dash
 frozenlist==1.4.1
     # via
@@ -232,7 +232,7 @@ pyerfa==2.0.0.3
     # via astropy
 pygelf==0.4.2
     # via katsdpservices
-pyjwt==2.8.0
+pyjwt==2.9.0
     # via katdal
 pymesos==0.3.15
     # via katsdpcontroller (setup.cfg)
@@ -255,7 +255,7 @@ referencing==0.33.0
     # via
     #   jsonschema
     #   jsonschema-specifications
-requests==2.32.2
+requests==2.32.4
     # via
     #   dash
     #   docker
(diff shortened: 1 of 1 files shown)


amishatishpatel (commented):
Before I explicitly hit Approve, Re: 
> Upgrades only the packages that does not require python upgrade

How did you converge on these packages (and versions)? I can't immediately make sense of dependabot's complaints.

amishatishpatel:
Also, if I could make a small complaint, your commit message of "upgradeables" isn't the most helpful.
- Usually we describe it as "bump versions" or "resolve dependabot alerts"
- I mean, I don't think version bumps are strictly upgrades either...

vincentseotlo:
Under "Security and quality -> View Dependabots alerts- >\<selected alert\>", Dependabot mentions the patched version of the affected package. Usually, updating pyproject.toml or requirements.txt with the suggested package does the trick. Other packages that are dependencies are also auto-updated by pre-commit.

In this push I could only update some packages. urllib3, for example, requires a newer Python version.

I have also updated the commit message to something more appropriate.

amishatishpatel (approved):
> Under "Security and quality -> View Dependabots alerts- ><selected alert>", Dependabot mentions the patched version of the affected package. Usually, updating pyproject.toml or requirements.txt with the suggested package does the trick. Other packages that are dependencies are also auto-updated by pre-commit.
> 
> In this push I could only update some packages. urllib3, for example, requires a newer Python version.
> 
> I have also updated the commit message to something more appropriate.

Fair points all around. I did go to various Release Notes and confirmed the versions you've bumped to in this PR.
- https://github.com/pallets/flask/releases
- https://github.com/jpadilla/pyjwt/releases
- https://github.com/ultrajson/ultrajson/releases
- https://github.com/urllib3/urllib3/blob/main/CHANGES.rst

If I'm honest though, and Re: My question on Mattermost/Python Support, I'm not actually sure/understanding how bumping versions to its "last-supported release" solves the dependabot issue.
- I'll Approve for now, mostly because sdp-jenkins reckons all is well, but 
- Perhaps chat with @bmerry during tomorrow's check-in on how this should proceed in future 
  - i.e. when patched version does not support target Python version

PR #836, opened 2026-05-19
Bypass _wait_rx_device_status for v engine

EatonEmmerich:
For NGC-1922 we need to start the v engine for qualification, currently qualification injects the data_timeout to the configuration which the v engine doesn't support.
This bypasses awaiting the data for the v engine rx status to update since the v engine only updates it after capture_start.

### DIFF
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -868,6 +868,8 @@ class SubarrayProduct:
         missing: Set[str] = set()  # Task names we're still waiting for
         future = asyncio.get_running_loop().create_future()
         for task in self.physical_graph:
+            if task.logical_node.task_type == 'v':
+                continue
             if isinstance(task, tasks.ProductPhysicalTaskMixin):
                 sensor = self.sdp_controller.sensors.get(f"{task.name}.rx.device-status")
                 if sensor is not None and sensor.status != Sensor.Status.NOMINAL:
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -868,7 +868,7 @@ class SubarrayProduct:
         missing: Set[str] = set()  # Task names we're still waiting for
         future = asyncio.get_running_loop().create_future()
         for task in self.physical_graph:
-            if task.logical_node.task_type == 'v':
+            if task.logical_node.task_type == "v":
                 continue
             if isinstance(task, tasks.ProductPhysicalTaskMixin):
                 sensor = self.sdp_controller.sensors.get(f"{task.name}.rx.device-status")
--- a/src/katsdpcontroller/product_controller.py
+++ b/src/katsdpcontroller/product_controller.py
@@ -869,6 +869,9 @@ class SubarrayProduct:
         future = asyncio.get_running_loop().create_future()
         for task in self.physical_graph:
             if isinstance(task, tasks.ProductPhysicalTaskMixin):
+                # V Engine currently doesn't support this check, as it only starts after the
+                # capture-start signal.
+                # TODO(NGC-1977): Improve this.
                 if task.task_type == "v":
                     continue
                 sensor = self.sdp_controller.sensors.get(f"{task.name}.rx.device-status")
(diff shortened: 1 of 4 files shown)


src/katsdpcontroller/product_controller.py
```diff
@@ -869,6 +869,11 @@ def observer(sensor, reading):
         future = asyncio.get_running_loop().create_future()
         for task in self.physical_graph:
             if isinstance(task, tasks.ProductPhysicalTaskMixin):
+                # V Engine currently doesn't support this check, as it only starts after the
+                # capture-start signal.
+                # TODO(NGC-1977): Improve this.
+                if task.task_type == "v":
```
bmerry: As discussed on MM, it should work if we just do

```suggestion
                if task.logical_node.task_type == "v":
```

and then revert the changes to `tasks.py` (and optionally move the `logical_node: ProductPhysicalTask` line from ProductPhysicalTask and ProductFakePhysicalTask down to ProductPhysicalTaskMixin).

bmerry (changes requested):

src/katsdpcontroller/scheduler.py
bmerry, line 1: Please revert the changes you made on this file. The scheduler exists at a lower layer of abstraction than katsdpcontroller.tasks, and should not import from it (this module is a general-purpose scheduler, while katsdpcontroller.tasks specialises for the way we do things on the telescope).

Where it would make sense to modify this file is to add `logical_node: LogicalTask` in PhysicalTask, as already appears in FakePhysicalTask.

src/katsdpcontroller/tasks.py
```diff
@@ -363,6 +363,7 @@ class ProductPhysicalTaskMixin(scheduler.PhysicalNode):
     """
 
     logical_node: ProductLogicalTask
+    name: str
```
bmerry: I don't see that adding anything: PhysicalNode already has a `name` of type `str`.

src/katsdpcontroller/tasks.py
```diff
@@ -686,8 +687,6 @@ class ProductPhysicalTask(ConfigMixin, ProductPhysicalTaskMixin, scheduler.Physi
       a port called ``prometheus``.
     """
 
-    logical_node: ProductLogicalTask
```
bmerry, line 686: If you remove this one and move it to the mixin class, remove it from ProductFakePhysicalTask too.

bmerry (changes requested):

src/katsdpcontroller/tasks.py
EatonEmmerich: Mypy requires this since the scheduler.FakePhysicalTask also defines a attribute with the same name but a different type than the ProductPhysicalTaskMixin:

```
src/katsdpcontroller/tasks.py:896: error: Definition of "logical_node" in base class "ProductPhysicalTaskMixin" is incompatible with definition in base class "FakePhysicalTask"  [misc]
Found 1 error in 1 file (checked 37 source files)
```
Since the python class is just a smash of everything together I agree that explicit here is more suitable.

src/katsdpcontroller/tasks.py
EatonEmmerich: the name in PhysicalNode is not typehinted though, I can typehint it there instead?

bmerry: True, it isn't. But I think the proper fix is to go through scheduler.py and give it proper type hints, which is beyond the scope of this PR. I've been adding type hints with new code, but I don't think we should type-hint `name` here (unless omitting it causes errors) because it's the "wrong" place to do it in the long term (it should be in the base class). Note that the hint on `logical_node` is different because it's overriding the type from the base class.

src/katsdpcontroller/tasks.py
```diff
@@ -370,6 +373,7 @@ def __init__(
         capture_block_id: str,
     ) -> None:
         # Turn .status into a property that updates a sensor
+        assert isinstance(self.logical_node, ProductLogicalTask)
```
bmerry: Is this needed, now that you have move the type declaration into this here?

src/katsdpcontroller/tasks.py
bmerry: Ah, I see, it's because the mixin inherits from PhysicalTask but not FakePhysicalTask. The fun and mess of diamond inheritance. Ok, I guess leave it as is.

bmerry (changes requested):

bmerry (approved):

PR #837, opened 2026-05-21
Copy telstate address into postprocess telstate node

richarms:
Fix continuum postprocessing in `katsdpcontroller` by copying the realtime telstate node's `address` onto the postprocess telstate external node.

The postprocess graph reuses telstate by creating a fake external node and copying over connection details from the realtime telstate task. It was copying `host` and `ports`, but not `address`. Batch endpoint resolution later expects `address`, and this left the telstate endpoint host as `None`, causing postprocessing task launch to fail while formatting task arguments.


KimMcAlpine (approved):
Yay. So glad this is fixed

PR #838, opened 2026-05-26
Improve error message when no usable offers received

bmerry:
Closes NGC-1937.

### DIFF
--- a/src/katsdpcontroller/diagnose_insufficient.py
+++ b/src/katsdpcontroller/diagnose_insufficient.py
@@ -152,6 +152,11 @@ class InsufficientRequesterVolume(InsufficientRequester):
         return f"{self.task.name} (volume {self.request.name})"
 
 
+class NoAgentsError(InsufficientResourcesError):
+    def __str__(self) -> str:
+        return "No usable (non-maintenance) agents were found"
+
+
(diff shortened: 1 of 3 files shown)


src/katsdpcontroller/diagnose_insufficient.py
```diff
@@ -538,6 +543,8 @@ def diagnose_insufficient(
         include non-tasks, which will be ignored.
     """
     with decimal.localcontext(DECIMAL_CONTEXT):
+        if not agents:
+            raise NoAgentsError()
```
EatonEmmerich, line 547: Why don't we simply add this check to _check_no_valid_agent instead? (`diagnose_isufficient.py:501`) where we loop the agents and raise a TaskNoValidAgentError

bmerry: I guess we could add these two lines there instead of here. Is there any advantage to doing it there though? For me doing it here makes it clear that we're first checking for a global condition (no agents at all) before digging into more local conditions (one specific task is unrunnable).

EatonEmmerich: I see, This is less confusing actually, :ok_hand:

EatonEmmerich (approved):

PR #842, opened 2026-06-12
NGC-1815: Determine resource requirements for V Engine

vincentseotlo:
In stead of reserving all memory on the GPU, we simply reserve what is needed, as detailed in 
https://skaafrica.atlassian.net/browse/NGC-1815


ludwigschwardt:
Just made the title nicer to match your branch name 😊 GitHub kinda mangles it...

src/katsdpcontroller/generator.py
```diff
@@ -1665,10 +1665,10 @@ def _make_vgpu(
     vgpu.interfaces[0].bandwidth_out = stream.data_rate() / n_engines
 
     vgpu.gpus = [scheduler.GPURequest()]
-    # TODO: To ensure vgpu doesn't share the GPU with anything else.
-    # Revisit once vgpu is complete.
+    # NOTE: Please see https://skaafrica.atlassian.net/browse/NGC-1815,
+    # on how mem formula is derived.
     vgpu.gpus[0].compute = 1.0
```
amishatishpatel: The original `TODO:` still holds true because `vgpu.gpus[0].compute` is still `1.0`.

This `NOTE:` belongs just above line 1671, but see my comment on the ticket Re: What is meant to be addressed in this PR.

amishatishpatel (changes requested):

src/katsdpcontroller/generator.py
vincentseotlo: Got it, I'll revert the comment and add mine just above the new change.

amishatishpatel (approved):
A reminder to keep the ticket open, maybe just unassign yourself.

src/katsdpcontroller/generator.py
```diff
@@ -1668,7 +1668,10 @@ def _make_vgpu(
     # TODO: To ensure vgpu doesn't share the GPU with anything else.
     # Revisit once vgpu is complete.
     vgpu.gpus[0].compute = 1.0
-    vgpu.gpus[0].mem = _mb(6e9)  # TODO: 6 GB for now; revisit once vgpu is complete
+
+    # NOTE: Please see https://skaafrica.atlassian.net/browse/NGC-1815,
+    # on how mem formula is derived.
```
bmerry: Rather put the measurements you obtained in the code here. katgpucbf is a public repository while the Jira is private. It doesn't need to be very detailed e.g. just mention observed GPU usage for bandwidths of 32 MHz, 64 MHz and 128 MHz.

bmerry (changes requested):

src/katsdpcontroller/generator.py
```diff
@@ -1665,12 +1665,15 @@ def _make_vgpu(
     vgpu.interfaces[0].bandwidth_out = stream.data_rate() / n_engines
 
     vgpu.gpus = [scheduler.GPURequest()]
-    # TODO: To ensure vgpu doesn't share the GPU with anything else.
-    # Revisit once vgpu is complete.
-    vgpu.gpus[0].compute = 1.0
+    # NOTE: Modeled estimate of the required GPU usage to accommodate the observed GPU usage
+    # for bandwidths of 32 MHz, 64 MHz, and 128 MHz. The observed GPU usage percentages
+    # are 45%, 29%, and 15%, respectively.
```
bmerry: Should it be the other way around (15%, 29%, and 45%)? The way you've written it, more bandwidth uses less CPU, which isn't what I would expect.

src/katsdpcontroller/generator.py
```diff
@@ -1665,12 +1665,15 @@ def _make_vgpu(
...
+    vgpu.gpus[0].compute = (0.303e-6 * stream.bandwidth + 10) / 100
+
+    # NOTE: Ensure there is enough GPU memory to accommodate the observed GPU usage
+    # for bandwidths of 32 MHz, 64 MHz, and 128 MHz. The observed GPU usage is 6152 MiB,
+    # 3956 MiB, and 2964 MiB, respectively.
```
bmerry: Again, you've listed these in reverse order.

src/katsdpcontroller/generator.py
```diff
@@ -1665,10 +1665,16 @@ def _make_vgpu(
     vgpu.interfaces[0].bandwidth_out = stream.data_rate() / n_engines
 
     vgpu.gpus = [scheduler.GPURequest()]
-    # TODO: To ensure vgpu doesn't share the GPU with anything else.
-    # Revisit once vgpu is complete.
-    vgpu.gpus[0].compute = 1.0
-    vgpu.gpus[0].mem = _mb(6e9)  # TODO: 6 GB for now; revisit once vgpu is complete
+    # NOTE: Modeled estimate of the required GPU usage to accommodate the observed GPU usage
```
bmerry:
```suggestion
    # Modeled estimate of the required GPU usage to accommodate the observed GPU usage
```

Comments are already "Notes", so unless a comment is unusually note-worthy (e.g. warnings to the developer about a risk, rather than just an explanation of the code) I wouldn't put that in.

src/katsdpcontroller/generator.py
```diff
@@ -1665,10 +1665,16 @@ def _make_vgpu(
...
+    # for bandwidths of 32 MHz, 64 MHz, and 128 MHz. The observed GPU usage percentages
+    # are 45%, 29%, and 15%, respectively.
+    vgpu.gpus[0].compute = (0.303e-6 * stream.bandwidth + 10) / 100
+
+    # NOTE: Ensure there is enough GPU memory to accommodate the observed GPU usage
```
bmerry:
```suggestion
    # Ensure there is enough GPU memory to accommodate the observed GPU usage
```

bmerry (changes requested):

src/katsdpcontroller/generator.py
vincentseotlo: True, 

I've tabulated the data. Thanks

src/katsdpcontroller/generator.py
vincentseotlo: Fixed.

vincentseotlo:
I'll be doing a run to verify the correctness of the models.

src/katsdpcontroller/generator.py
```diff
@@ -1638,9 +1638,23 @@ def _make_vgpu(
 
     for ss in stream_sensors:
         g.graph["stream_sensors"].add(ss)
+    # Observed data
+    # +-------------------------------------+
+    # |Narrowband         |16   |8    |4    |
+    # +-------------------------------------+
+    # |Bandwidth  (Mhz)   |32   |64   |128  |
```
bmerry:
```suggestion
    # |Bandwidth  (MHz)   |32   |64   |128  |
```

src/katsdpcontroller/generator.py
```diff
@@ -1638,9 +1638,23 @@ def _make_vgpu(
...
+    # |vgpu res memory(M) |1200 |1299 |1629 |
+    # |vgpu gpu usage (%) |15   |29   |45   |
+    # |vgpu gpu memory(M) |2964 |3956 |6152 |
+    # +-------------------------------------+
+
+    # Estimate required cpu resources
+    vgpu.cpus = (0.315e-5 * stream.bandwidth + 5) / 100
+
+    # Estimate required host memory
+    vgpu.mem = round(4.6e-5 * stream.bandwidth) + 1435
```
bmerry: When I plug in bandwidth=128e6 to this formula I get 7323 MB. We do definitely want some padding since this is a hard limit, but we don't need 5x padding. Maybe 2x would be reasonable.

src/katsdpcontroller/generator.py
```diff
@@ -1638,9 +1638,23 @@ def _make_vgpu(
...
+    # +-------------------------------------+
+    # |Bandwidth  (Mhz)   |32   |64   |128  |
+    # |vgpu cpu usage (%) |12   |21   |42   |
+    # |vgpu res memory(M) |1200 |1299 |1629 |
+    # |vgpu gpu usage (%) |15   |29   |45   |
+    # |vgpu gpu memory(M) |2964 |3956 |6152 |
+    # +-------------------------------------+
+
+    # Estimate required cpu resources
+    vgpu.cpus = (0.315e-5 * stream.bandwidth + 5) / 100
```
bmerry: When I plug in the 3 example bandwidth numbers, I get 1.058, 2.066 and 4.082, which seems much higher than the numbers in your table. I think you have a factor-of-ten error in the scaling coefficient.

I'd also suggest clamping this with `min(..., 1.0)`: it's single-threaded so it won't be able to use more than 1 CPU anyway.

src/katsdpcontroller/generator.py
```diff
@@ -1663,12 +1677,21 @@ def _make_vgpu(
...
     vgpu.interfaces[0].bandwidth_out = stream.data_rate() / n_engines
-
     vgpu.gpus = [scheduler.GPURequest()]
-    # TODO: To ensure vgpu doesn't share the GPU with anything else.
-    # Revisit once vgpu is complete.
-    vgpu.gpus[0].compute = 1.0
-    vgpu.gpus[0].mem = _mb(6e9)  # TODO: 6 GB for now; revisit once vgpu is complete
+
+    # Estimate required GPU usage
+    vgpu.gpus[0].compute = (0.303e-5 * stream.bandwidth + 17) / 100
```
bmerry: When I plug in the bandwidth numbers I'm getting values bigger than 1 here, which means it probably won't even launch.

src/katsdpcontroller/generator.py
```diff
@@ -1600,9 +1600,23 @@ def _make_vgpu(
...
+    # +-------------------------------------+
+    # |Narrowband         |16   |8    |4    |
+    # +-------------------------------------+
+    # |Bandwidth  (MHz)   |32   |64   |128  |
+    # |vgpu cpu usage (%) |12   |21   |42   |
+    # |vgpu res memory(M) |1200 |1299 |1629 |
+    # |vgpu gpu usage (%) |15   |29   |45   |
+    # |vgpu gpu memory(M) |2964 |3956 |6152 |
+    # +-------------------------------------+
+    bandwidth = stream.bandwidth * 1e-6
```
bmerry: Can you call this `bandwidth_mhz` to make it extra-clear to the reader that this is in non-base units.

bmerry (changes requested):
Formulae look sensible now. Just a request to rename the variable to be more explicit about units.

bmerry (approved):

PR #843, opened 2026-06-19
Add VLBI VDIF recording and metadata post-processing

richarms:
This adds product configuration and controller support for VLBI VDIF recording and postprocessing metadata generation from a gpucbf V-engine.

This branch introduces the `sdp.vdif` output stream type in product-config schema, wires it through product configuration, and generates the corresponding capture-time `katsdpvlbi` recorder task. The post-processing task is scheduled with the capture block, receives telstate context, can run  in `antab`, `pass_through`, or `disabled` mode, and is colocated with the corresponding VLBI recorder. Implements version 4.8 of the product-configure schema. Tested end-to-end in the lab.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -123,6 +123,7 @@ IMAGES = frozenset(
         "katsdpdatawriter",
         "katsdpmetawriter",
         "katsdptelstate",
+        "katsdpvlbi",
     ]
 )
 #: Number of bytes used by spectral imager per visibility
@@ -3042,6 +3043,8 @@ def build_logical_graph(
         )
     for vgpu_stream in configuration.by_class(product_config.GpucbfTiedArrayResampledVoltageStream):
         _make_vgpu(g, configuration, vgpu_stream)
+    for vdif_stream in configuration.by_class(product_config.VdifStream):
+        _make_vlbi(g, configuration, vdif_stream)
 
     # Pair up spectral and continuum L0 outputs
     l0_done = set()
@@ -3587,6 +3590,72 @@ def _make_spectral_imager_report(
     for node in spectral_nodes:
         g.add_edge(report, node, depends_finished=True, depends_finished_critical=False)
     return report
+
+
+def _make_vlbi(
+    g: networkx.MultiDiGraph, configuration: Configuration, stream: product_config.VdifStream
+) -> scheduler.LogicalNode:
+    """Create a capture-time VLBI recorder task for a VDIF stream."""
+    source_stream = stream.source_stream
+    if not isinstance(source_stream, product_config.GpucbfTiedArrayResampledVoltageStream):
+        raise NotImplementedError(
+            "sdp.vdif capture currently requires gpucbf.tied_array_resampled_voltage"
+        )
+    source_multicast = find_node(g, "multicast." + source_stream.name)
+    assert isinstance(source_multicast, LogicalMulticast)
+    task = ProductLogicalTask(f"vlbi.{stream.name}", streams=[stream])
+    task.subsystem = "sdp"
+    task.cpus = 2.0
+    task.mem = 8 * 1024
+    size_bytes = getattr(source_stream, "size", None)
+    if size_bytes is not None:
+        task.disk = _mb(1024 * size_bytes + 1024)
+    task.volumes = [DATA_VOL]
+    task.image = "katsdpvlbi"
+    task.katsdpservices_config = False
+    task.katsdpservices_logging = False
+    task.pass_telstate = False
+    task.metadata_katcp_sensors = False
+    task.transitions = CAPTURE_TRANSITIONS
+    task.ports = ["port"]
+    task.wait_ports = ["port"]
+    payload_bytes = (defaults.VGPU_SAMPLES_PER_FRAME * source_stream.bits_per_sample) // 8
+    frame_bytes = 32 + payload_bytes
+    n_threads = source_stream.n_chans * len(source_stream.pols)
+    j5a_mode = f"vdif_{payload_bytes}-{frame_bytes}-{source_stream.bits_per_sample}-{n_threads}"
+    command_lines = [
+        'endpoint="$1"',
+        'katcp_port="$2"',
+        'export J5A_NETPORT="${{endpoint/:/@}}"',
+        "export J5A_PROTOCOL=udps",
+        f"export J5A_MODE={j5a_mode}",
+        'export KATCP_PORT="$katcp_port"',
+        "export KATCP_ENABLE=true",
+        "export AUTOSTART_RECORD=false",
+        f"export DISK_PATHS={escape_format(DATA_VOL.container_path)}",
+    ]
+    if frame_bytes + 8 > 1500:
+        command_lines.append("export J5A_MTU=9000")
+        command_lines.append("export J5A_BUFF_SND=268435456")
+    command_lines.append("exec /usr/local/bin/entrypoint.sh")
+    task.command = [
+        "bash",
+        "-ceu",
+        "\n".join(command_lines),
+        "_",
+        f"{{endpoints[multicast.{source_stream.name}_vdif]}}",
+        "{ports[port]}",
+    ]
+    g.add_node(task)
+    g.add_edge(
+        task,
+        source_multicast,
+        port="vdif",
+        depends_resolve=True,
+        depends_init=True,
+        depends_ready=True,
+    )
+    return task
 
 
 async def build_postprocess_logical_graph(
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -124,6 +124,7 @@ IMAGES = frozenset(
         "katsdpmetawriter",
         "katsdptelstate",
         "katsdpvlbi",
+        "vlbimeta",
     ]
 )
 #: Number of bytes used by spectral imager per visibility
@@ -3658,6 +3659,35 @@ def _make_vlbi(
     return task
 
 
+def _make_vlbimeta(
+    g: networkx.MultiDiGraph,
+    configuration: Configuration,
+    capture_block_id: str,
+    stream: product_config.VdifStream,
+) -> scheduler.LogicalNode:
+    """Create a post-processing task for VLBI metadata extraction."""
+    task = ProductLogicalTask(f"vlbimeta.{stream.name}", streams=[stream])
+    task.subsystem = "sdp"
+    task.cpus = 1.0
+    task.mem = 4 * 1024
+    task.volumes = [DATA_VOL]
+    task.image = "vlbimeta"
+    task.katsdpservices_config = False
+    task.katsdpservices_logging = False
+    task.pass_telstate = True
+    task.metadata_katcp_sensors = False
+    task.command = [
+        "vlbimeta.py",
+        escape_format(DATA_VOL.container_path),
+        escape_format(capture_block_id),
+        escape_format(stream.name),
+        "--mode",
+        configuration.options.vlbimeta.mode,
+    ]
+    g.add_node(task)
+    return task
+
+
 async def build_postprocess_logical_graph(
     configuration: Configuration,
     capture_block_id: str,
@@ -3680,6 +3710,10 @@ async def build_postprocess_logical_graph(
             g, configuration, capture_block_id, cstream, telstate, telstate_endpoint, target_mapper
         )
 
+    for vdif_stream in configuration.by_class(product_config.VdifStream):
+        if configuration.options.vlbimeta.mode != "disabled":
+            _make_vlbimeta(g, configuration, capture_block_id, vdif_stream)
+
     # Note: this must only be run after all the sdp.continuum_image nodes have
     # been created, because spectral imager nodes depend on continuum imager
     # nodes.
@@ -3700,6 +3734,10 @@ async def build_postprocess_logical_graph(
             seen.add(node.name)
             assert node.image in IMAGES, f"{node.image} missing from IMAGES"
             # Connect every task to telstate
+            if node.pass_telstate:
+                node.command.extend(
+                    ["--telstate", "{endpoints[telstate_telstate]}", "--name", node.name]
+                )
             g.add_edge(node, telstate_node, port="telstate", depends_ready=True, depends_kill=True)
 
     return g
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -3602,6 +3602,7 @@ def _make_vlbi(
         raise NotImplementedError(
             "sdp.vdif capture currently requires gpucbf.tied_array_resampled_voltage"
         )
+    develop = configuration.options.develop
     source_multicast = find_node(g, "multicast." + source_stream.name)
     assert isinstance(source_multicast, LogicalMulticast)
     task = ProductLogicalTask(f"vlbi.{stream.name}", streams=[stream])
@@ -3611,6 +3612,15 @@ def _make_vlbi(
     size_bytes = getattr(source_stream, "size", None)
     if size_bytes is not None:
         task.disk = _mb(1024 * size_bytes + 1024)
+    task.interfaces = [
+        scheduler.InterfaceRequest(
+            "cbf",
+            affinity=not develop.disable_ibverbs,
+            infiniband=not develop.disable_ibverbs,
+            multicast_in={source_stream.name},
+        )
+    ]
+    task.interfaces[0].bandwidth_in = source_stream.data_rate()
     task.volumes = [DATA_VOL]
     task.image = "katsdpvlbi"
     task.katsdpservices_config = False
@@ -3628,8 +3638,9 @@ def _make_vlbi(
         'endpoint="$1"',
         'katcp_port="$2"',
         'export J5A_NETPORT="${{endpoint/:/@}}"',
-        "export J5A_PROTOCOL=udps",
+        f"export J5A_PROTOCOL={develop.vlbi_recorder_protocol}",
         f"export J5A_MODE={j5a_mode}",
+        'export J5A_CBF_INTERFACE="{interfaces[cbf].name}"',
         'export KATCP_PORT="$katcp_port"',
         "export KATCP_ENABLE=true",
         "export AUTOSTART_RECORD=false",
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -3675,6 +3675,7 @@ def _make_vlbimeta(
     configuration: Configuration,
     capture_block_id: str,
     stream: product_config.VdifStream,
+    dataset_stream: product_config.VisStream,
 ) -> scheduler.LogicalNode:
     """Create a post-processing task for VLBI metadata extraction."""
     task = ProductLogicalTask(f"vlbimeta.{stream.name}", streams=[stream])
@@ -3692,6 +3693,8 @@ def _make_vlbimeta(
         escape_format(DATA_VOL.container_path),
         escape_format(capture_block_id),
         escape_format(stream.name),
+        "--dataset-stream-name",
+        escape_format(dataset_stream.name),
         "--mode",
         configuration.options.vlbimeta.mode,
     ]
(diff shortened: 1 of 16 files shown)


src/katsdpcontroller/schemas/product_config.yaml.j2
```diff
@@ -733,11 +751,28 @@ properties:
               less_resources:
                 type: boolean
                 default: false
+              vlbi_recorder_protocol:
```
bmerry, line 776: I don't see this in the schema document. And once you've added it, it needs to be gated on `if version >= "4.8"`.

src/katsdpcontroller/schemas/product_config.yaml.j2
```diff
@@ -733,11 +751,28 @@ properties:
...
+                type: string
```
bmerry: This doesn't hurt, but it's redundant since the `enum` limits it to particular strings anyway.

src/katsdpcontroller/schemas/product_config.yaml.j2
```diff
@@ -733,11 +751,28 @@ properties:
...
+                enum:
+                  - udps
+                  - udpsnor
+                default: udps
             additionalProperties: false
 {% endif %}
           - type: boolean
             default: false
         default: false
+      vlbimeta:
```
bmerry, line 788: Needs to be gated on `if version >= "4.8"`

src/katsdpcontroller/product_config.py
```diff
@@ -252,26 +252,37 @@ def from_config(cls, config: Mapping[str, Any]) -> "ServiceOverride":
...
         less_resources: bool = False,
         data_timeout: float = 0.0,
+        vlbi_recorder_protocol: str = "udps",
     ) -> None:
+        if vlbi_recorder_protocol not in self.VALID_VLBI_RECORDER_PROTOCOLS:
+            valid = ", ".join(sorted(self.VALID_VLBI_RECORDER_PROTOCOLS))
+            raise ValueError(
+                f"Unknown VLBI recorder protocol {vlbi_recorder_protocol!r}. "
+                f"Expected one of: {valid}"
+            )
```
bmerry: This isn't necessary, because the jsonschema validator will already validate it.

src/katsdpcontroller/product_config.py
```diff
@@ -283,11 +294,26 @@ def from_bool(cls, opt: bool) -> "DevelopOptions":
         )
 
 
+class VlbimetaOptions:
+    VALID_MODES = frozenset({"antab", "pass_through", "disabled"})
+
+    def __init__(self, *, mode: str = "antab") -> None:
+        if mode not in self.VALID_MODES:
+            valid_modes = ", ".join(sorted(self.VALID_MODES))
+            raise ValueError(f"Unknown vlbimeta mode {mode!r}. Expected one of: {valid_modes}")
```
bmerry: This validation isn't necessary: the jsonschema already enforces it.

src/katsdpcontroller/product_config.py
```diff
@@ -1878,6 +1906,52 @@ def from_config(
         )
 
 
+class VdifStream(Stream):
+    """VDIF-encapsulated voltages from a tied-array stream."""
+
+    stream_type: ClassVar[str] = "sdp.vdif"
+    _valid_src_types: ClassVar[_ValidTypes] = {
+        "gpucbf.tied_array_resampled_voltage",
```
bmerry:
```suggestion
        "cbf.tied_array_resampled_voltage",
```

src/katsdpcontroller/product_config.py
```diff
@@ -1878,6 +1906,52 @@ def from_config(
...
+
+    def __init__(self, name: str, src_streams: Sequence[Stream]) -> None:
+        if len(src_streams) != 1:
+            raise ValueError("Exactly one tied-array resampled voltage source is required")
+        super().__init__(name, src_streams)
+
+    @property
+    def source_stream(
+        self,
+    ) -> Union[GpucbfTiedArrayResampledVoltageStream, SimTiedArrayChannelisedVoltageStream]:
```
bmerry: You'll need to create a CbfTiedArrayResampledVoltageStream type to represent the cbf.tied_array_resampled_voltage type.

src/katsdpcontroller/product_config.py
```diff
@@ -1878,6 +1906,52 @@ def from_config(
         )
 
 
+class VdifStream(Stream):
+    """VDIF-encapsulated voltages from a tied-array stream."""
+
+    stream_type: ClassVar[str] = "sdp.vdif"
+    _valid_src_types: ClassVar[_ValidTypes] = {
+        "gpucbf.tied_array_resampled_voltage",
+        "sim.cbf.tied_array_channelised_voltage",
```
bmerry: That's not allowed according to the schema. And I don't know how you'd implement it since it's a totally different type of stream.

test/test_product_config.py
```diff
@@ -1864,11 +1943,13 @@ def mock_time(self, mocker) -> None:
     def test_upgrade_v3(self, config_v3: Dict[str, Any], config: Dict[str, Any]) -> None:
         upgraded = product_config._upgrade(config_v3)
         config["outputs"]["m002h"]["sync_time"] = 123456789.0  # Added by _upgrade
+        config["version"] = "4.8"
```
bmerry: Just update the `config` fixture, rather than "fixing" it each time it's used in a test.

test/test_product_config.py
```diff
@@ -1985,6 +2066,44 @@ async def test_mixed_bands(self, config: Dict[str, Any]) -> None:
         with pytest.raises(ValueError, match="Only a single band is supported, found 'l', 'u'"):
             await Configuration.from_config(config)
 
+    async def test_vlbi_recorder_uses_cbf_multicast_interface(self) -> None:
```
bmerry: Rather create a test_generator.py and put these tests there.

test/utils.py
```diff
@@ -348,6 +352,10 @@
             "n_chans": 2,
             "pols": ["x", "y"],
             "station_id": "me"
+        },
+        "sdp_vdif": {
```
bmerry: Please don't add an SDP stream to CONFIG_CBF_ONLY.

test/utils.py
```diff
@@ -130,6 +130,10 @@
             "pols": ["x", "y"],
             "station_id": "me"
         },
+        "sdp_vdif": {
+            "type": "sdp.vdif",
+            "src_streams": ["gpucbf_tied_array_resampled_voltage"]
```
bmerry: That's not a valid stream type for sdp.vdif.

bmerry (changes requested):

test/utils.py
ludwigschwardt: I'm now thoroughly confused... From the product config schema:
```jinja2
  stream_name:
    type: string
    pattern: "{{ stream_name_pattern }}"
  stream_type:
    type: string
    pattern: "^[A-Za-z0-9_.]+$"
  src_streams:
    type: array
    items:
      $ref: "#/definitions/stream_name"
```
According to this the `src_streams` key contains stream *names*, not *types*.

1. Is `gpucbf_tied_array_resampled_voltage` a stream name?
2. Is `cbf.tied_array_resampled_voltage` a stream type?
3. What about `gpucbf.tied_array_resampled_voltage`? Does the dot essentially indicate a type (since names can't have dots)?
4. Just above this code snippet, there is a definition for the `gpucbf_tied_array_resampled_voltage` stream that seems to mesh with this.
5. When would you use `cbf` and when `gpucbf`?
6. Or are you referring to `"sdp.vdif"`? 😅

richarms:
I've pushed follow-up changes to adress review comments:

- Added `4.8` gating for the new VLBI config options in product config schema.
- Removed duplicate Python-side enum validation for `vlbi_recorder_protocol` and `vlbimeta.mode`; these are now left to jsonschema validation.
- Added `cbf.tied_array_resampled_voltage` as the CBF-facing stream type for vlbi. Changed `sdp.vdif` to require `cbf.tied_array_resampled_voltage` as its source.
- Removed the SDP stream from `CONFIG_CBF_ONLY`.
- Moved the generator-specific VLBI tests into `test_generator.py`.

src/katsdpcontroller/schemas/product_config.yaml.j2
```diff
@@ -243,10 +244,16 @@ properties:
 {% endif %}
 {% if version >= "4.7" %}
               - gpucbf.tied_array_resampled_voltage
+{% endif %}
+{% if version >= "4.8" %}
+              - cbf.tied_array_resampled_voltage
```
bmerry: This should be in input types, not output types.

src/katsdpcontroller/schemas/product_config.yaml.j2
```diff
@@ -311,6 +318,32 @@ properties:
...
+                type:
+                  const: cbf.tied_array_resampled_voltage
+            then:
+              properties:
+                type: {}
+                src_streams:
+                  $ref: "#/definitions/singleton"
+              required:
+                - src_streams
+              additionalProperties: false
```
bmerry: Move to input types. You'll probably also need to follow the pattern of the other `cbf.*` input types for it to work.

src/katsdpcontroller/schemas/product_config.yaml.j2
bmerry: It's still not in the schema document, nor in the signed M1000-0001-009 v4. So you'll probably need an ECP for this.

src/katsdpcontroller/generator.py
```diff
@@ -3521,6 +3593,112 @@ def _make_spectral_imager_report(
...
+    task.volumes = [DATA_VOL]
+    task.image = "katsdpvlbi"
+    task.katsdpservices_config = False
+    task.katsdpservices_logging = False
+    task.pass_telstate = False
+    task.metadata_katcp_sensors = False
+    task.transitions = CAPTURE_TRANSITIONS
+    task.ports = ["port"]
+    task.wait_ports = ["port"]
+    payload_bytes = (defaults.VGPU_SAMPLES_PER_FRAME * source_stream.bits_per_sample) // 8
```
bmerry: I guess this works for now, but I'll file a ticket to add a sensor so that you can actually get the live value (we've also been thinking about the sensor for internal use anyway).

src/katsdpcontroller/product_config.py
```diff
@@ -1492,6 +1507,54 @@ def from_config(
         )
 
 
+class CbfTiedArrayResampledVoltageStream(Stream):
+    """VDIF-ready tied-array-resampled-voltage stream from CBF."""
+
+    stream_type: ClassVar[str] = "cbf.tied_array_resampled_voltage"
+    _valid_src_types: ClassVar[_ValidTypes] = {"gpucbf.tied_array_resampled_voltage"}
```
bmerry:
```suggestion
    _valid_src_types: ClassVar[_ValidTypes] = {"cbf.tied_array_channelised_voltage"}
```
I see M1000-0000-009 incorrectly lists it at cbf.tied_array_resampled_voltage. That'll need fixing.

src/katsdpcontroller/schemas/product_config.yaml.j2
bmerry: It looks like this type was added to M1000-0001-009, but not to https://docs.google.com/document/d/15ZIyuBf4Vk4ESIEdKsQxdSBoLEwh85eyKeDJouLRAZY. Please can you add it to the latter.

src/katsdpcontroller/product_config.py
```diff
@@ -1492,6 +1507,54 @@ def from_config(
...
+
+    def __init__(self, name: str, src_streams: Sequence[Stream]) -> None:
+        if len(src_streams) != 1:
+            raise ValueError("Exactly one tied-array resampled voltage source is required")
+        if not isinstance(src_streams[0], GpucbfTiedArrayResampledVoltageStream):
+            raise ValueError("Source stream must be gpucbf.tied_array_resampled_voltage")
+        super().__init__(name, src_streams)
+
```
bmerry:
```suggestion
```

The schema should validate the number of src_streams, and `_validate` already checks the types against `_valid_src_types`.

src/katsdpcontroller/product_config.py
```diff
@@ -1492,6 +1507,54 @@ def from_config(
...
+    @property
+    def source_stream(self) -> GpucbfTiedArrayResampledVoltageStream:
+        return cast(GpucbfTiedArrayResampledVoltageStream, self.src_streams[0])
```
bmerry: I think there is some confusion here. Only the CBF product controller deals in gpucbf.* streams. The SDP product controller gets data from a "CBF" - whether that's implemented on SKARABs, GPUs or magic goblins.

It then queries CAM for sensors to find out things like the bandwidth or bits per sample.

A cbf.tied_array_resampled_voltage stream has two source streams, of type cbf.tied_array_channelised_voltage (one for each polarisation). And the point of a V-Engine is to resample to a different bandwidth, bit-depth, etc, so you can't simply get those values from the upstream stream. They need to come from sensors.

src/katsdpcontroller/product_config.py
```diff
@@ -1878,6 +1941,41 @@ def from_config(
...
+    """VDIF-encapsulated voltages from a tied-array stream."""
+
+    stream_type: ClassVar[str] = "sdp.vdif"
+    _valid_src_types: ClassVar[_ValidTypes] = {"cbf.tied_array_resampled_voltage"}
+
+    def __init__(self, name: str, src_streams: Sequence[Stream]) -> None:
+        if len(src_streams) != 1:
+            raise ValueError("Exactly one tied-array resampled voltage source is required")
+        super().__init__(name, src_streams)
+
```
bmerry:
```suggestion
```
The jsonschema should check this.

src/katsdpcontroller/tasks.py
```diff
@@ -334,6 +335,57 @@ def close(self):
         self.sensor.detach(self)
 
 
+class TelstateSensorHistoryObserver:
+    """Persist selected mirrored sensor readings into the active capture-block telstate view."""
```
bmerry: This looks like it's reinventing cam2telstate. @ludwigschwardt am I missing something?

bmerry (changes requested):

richarms:
Thanks for the detailed reviews. I have reworked the branch to implement your stream-model and schema suggestions.

- The VLBI additions are gated on product-config schema 4.8.
- `cbf.tied_array_resampled_voltage` is now an external input type following the `cbf.*` pattern.
- It requires exactly two `cbf.tied_array_channelised_voltage` source streams.
- Its bandwidth, channel count, sample depth and polarisation ordering come from CBF sensors obtained through CAM.
- sdp.vdif accepts only this external CBF stream as its source; the direct `gpucbf.*` paths have been removed.
- Redundant schema validation was removed.
- The shared product-config fixture was updated to 4.8, and generator-specific tests were moved to `test_generator.py`.
- Mean-power history is now handled by katsdpcam2telstate, rather than adding things to controller.

I also (successfully) ran the branch through in the lab.

src/katsdpcontroller/schemas/product_config.yaml.j2
bmerry: Still not addressed as far as I'm aware.

src/katsdpcontroller/generator.py
```diff
@@ -3522,6 +3593,111 @@ def _make_spectral_imager_report(
     return report
 
 
+def _make_vlbi(
+    g: networkx.MultiDiGraph, configuration: Configuration, stream: product_config.VdifStream
+) -> scheduler.LogicalNode:
```
bmerry: This is an odd place to put this function: all the other real-time nodes created by build_logical_graph are located above build_logical_graph.

src/katsdpcontroller/product_config.py
```diff
@@ -1492,6 +1508,65 @@ def from_config(
         )
 
 
+class CbfTiedArrayResampledVoltageStream(CbfStream, Stream):
+    """VDIF-ready tied-array-resampled-voltage stream from CBF."""
```
bmerry: What does "VDIF-ready" mean?

test/test_generator.py
```diff
@@ -0,0 +1,81 @@
+################################################################################
+# Copyright (c) 2013-2025, National Research Foundation (SARAO)
```
bmerry:
```suggestion
# Copyright (c) 2026, National Research Foundation (SARAO)
```
This is a new file and new code. See https://skaafrica.atlassian.net/wiki/spaces/DPP/pages/2938961934/Copyright+notices

test/test_product_config.py
```diff
@@ -1235,6 +1247,56 @@ def test_bad_n_chans(
...
+            ),
+            make_tied_array_channelised_voltage(
+                acv,
+                "vlbi_tied_array_channelised_voltage_0y",
+                yarl.URL("spead://239.2.4.0+3:7148"),
+            ),
+        ]
+        config = {
+            "type": "cbf.tied_array_resampled_voltage",
+            "url": "spead://239.2.5.0+3:7148",
```
bmerry: tied_array_resampled_voltage doesn't use SPEAD; it uses VDIF.

test/test_product_controller.py
```diff
@@ -535,6 +534,7 @@ def mock_katportalclient(self, mocker) -> None:
                 "cbf_1_i0_tied_array_channelised_voltage_0y_spectra_per_heap": 256,
                 "cbf_1_i0_tied_array_channelised_voltage_0y_n_chans_per_substream": 256,
                 "cbf_1_i0_tied_array_channelised_voltage_0y_beng_out_bits_per_sample": 8,
+                **VLBI_CBF_SENSOR_VALUES,
```
bmerry: This is going to replace some of the values above e.g. cbf_1_i0_adc_sample_rate. That could be very confusing if anyone ever tried to change them and it has no effect.

test/utils.py
```diff
@@ -253,8 +252,70 @@
...
+            },
+            "vlbi_tied_array_channelised_voltage_0y": {
+                "type": "cbf.tied_array_channelised_voltage",
+                "url": "spead://239.2.4.0+3:7148",
+                "src_streams": ["vlbi_antenna_channelised_voltage"],
+                "instrument_dev_name": "i0",
+            },
+            "cbf_tied_array_resampled_voltage": {
+                "type": "cbf.tied_array_resampled_voltage",
+                "url": "spead://239.2.5.0+3:7148",
```
bmerry: It's not SPEAD.

bmerry (changes requested):
Looking much better architecturally.

src/katsdpcontroller/schemas/product_config.yaml.j2
bmerry: I see you've added it, but it wasn't listed in the version history. I've added that now.

src/katsdpcontroller/generator.py
bmerry: I've filed NGC-2071 for this.

bmerry (commented):
Code changes look okay, but you've still got this vlbi_recorder_protocol in the implementation that's not defined in the ICD.

richarms:
Great thanks. I suggest that we not make that block this PR, and I'll follow this up as an ICD amendment

bmerry (approved):

test/test_generator.py
```diff
@@ -0,0 +1,88 @@
...
+#   https://opensource.org/licenses/BSD-3-Clause
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+"""Tests for :mod:`katsdpcontroller.generator`."""
```
ludwigschwardt, line 17: The first generator unit tests... Exciting!

test/utils.py
```diff
@@ -253,8 +252,70 @@
...
+                "type": "cbf.tied_array_channelised_voltage",
+                "url": "spead://239.2.4.0+3:7148",
+                "src_streams": ["vlbi_antenna_channelised_voltage"],
+                "instrument_dev_name": "i0",
+            },
+            "cbf_tied_array_resampled_voltage": {
+                "type": "cbf.tied_array_resampled_voltage",
+                "url": "vdif://239.2.5.0+3:7148",
+                "src_streams": [
+                    "vlbi_tied_array_channelised_voltage_0x",
```
ludwigschwardt, line 304: Is "vlbi" playing the role of "wide" (i.e. a CBF product name aka the old "instrument")?

ludwigschwardt (approved):
Looks good (although I confess that a lot of the machinery goes over my head at this stage!).

ludwigschwardt:
Is there an SPP1 ticket associated with this?

test/utils.py
richarms: Similar, but not quite -- `vlbi` is the stream-name prefix in the test fixture. The CBF product/old instrument name is explicitly `i0` here, through `instrument_dev_name`.

richarms:
> Is there an SPP1 ticket associated with this?

I made several tickets that relate to this under https://skaafrica.atlassian.net/browse/MKAIV-1635, which I will update.

PR #845, opened 2026-06-23
Allow for per-output send-enabled in xbgpu

amishatishpatel:
As per https://github.com/ska-sa/katgpucbf/pull/1061

Admittedly, I have **not yet** added logic for the `isinstance(stream, GpucbfBaselineCorrelationProductsStream)` (`--corrprod`) branch of logic. I figured that was not part of this ticket - and it can be done with a `?capture-start` (?).

Also, 
- Updated the product controller's `capture_start_stop` method to **not** allow V-engine src-streams to be disabled, and
- Tweaked a unit test or two to check this updated `?capture-stop` logic.

Contributes to: NGC-1990.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1438,10 +1438,26 @@ def _make_xbgpu(
                     "name": escape_format(stream.name),
                     "dst": f"{{endpoints_vector[multicast.{stream.name}_spead][{i}]}}",
                     "pol": stream.src_pol,
-                    "send_enabled": _has_downstream_dependencies(
-                        stream, vgpu_streams, check_by_name=True
-                    ),
                 }
+                if _has_downstream_dependencies(stream, vgpu_streams, check_by_name=True):
+                    output_config["send_enabled"] = True
+                    gpucbf_tacv_dst_mcast = list(
+                        filter(
+                            lambda dst_mcast: dst_mcast.name == f"multicast.{stream.name}",
+                            dst_multicasts,
+                        )
+                    )
+                    if len(gpucbf_tacv_dst_mcast) == 1:
+                        gpucbf_tacv_dst_mcast[0].initial_transmit_state = TransmitState.UP
+                    else:
+                        # Found more than one or none, which is unexpected
+                        raise ValueError(
+                            f"Expected exactly one multicast for stream {stream.name}, "
+                            "found {len(gpucbf_tacv_dst_mcast)}"
+                        )
+                else:
+                    output_config["send_enabled"] = False
+
                 if stream.dither is not None:
                     output_config["dither"] = stream.dither
                 xbgpu.command += [
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -963,6 +963,22 @@ def _make_fgpu(
     return fgpu_group
 
 
+def _has_downstream_dependencies(
+    src_stream: product_config.Stream,
+    dst_streams: Sequence[product_config.Stream],
+    check_by_name: bool = False,
+) -> bool:
+    """Check if `src_stream` is used by any of the `dst_streams`."""
+    for dst_stream in dst_streams:
+        if check_by_name:
+            if src_stream.name in [stream.name for stream in dst_stream.src_streams]:
+                return True
+        else:
+            if src_stream in dst_stream.src_streams:
+                return True
+    return False
+
+
 def _make_xbgpu(
     g: networkx.MultiDiGraph,
     configuration: Configuration,
@@ -1284,6 +1300,8 @@ def _make_xbgpu(
     recv_buffer = free_chunks * chunk_size
 
     task_names = []
+    # Needed for send_enabled command-line arg for beam data streams
+    vgpu_streams = configuration.by_class(product_config.GpucbfTiedArrayResampledVoltageStream)
     for i in range(n_engines):
         # One engine per section of the band
         xbgpu = ProductLogicalTask(f"xb.{base_name}.{i}", streams=streams, index=i)
@@ -1420,6 +1438,9 @@ def _make_xbgpu(
                     "name": escape_format(stream.name),
                     "dst": f"{{endpoints_vector[multicast.{stream.name}_spead][{i}]}}",
                     "pol": stream.src_pol,
+                    "send_enabled": _has_downstream_dependencies(
+                        stream, vgpu_streams, check_by_name=True
+                    ),
                 }
                 if stream.dither is not None:
                     output_config["dither"] = stream.dither
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -966,16 +966,11 @@ def _make_fgpu(
 def _has_downstream_dependencies(
     src_stream: product_config.Stream,
     dst_streams: Sequence[product_config.Stream],
-    check_by_name: bool = False,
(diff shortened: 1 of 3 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1421,6 +1439,25 @@ def get_baseline_index(a1, a2):
...
                     "pol": stream.src_pol,
                 }
+                if _has_downstream_dependencies(stream, vgpu_streams, check_by_name=True):
+                    output_config["send_enabled"] = True
+                    gpucbf_tacv_dst_mcast = list(
+                        filter(
+                            lambda dst_mcast: dst_mcast.name == f"multicast.{stream.name}",
+                            dst_multicasts,
+                        )
+                    )
```
bmerry: Rather than going looking for it, I think you can just change the iteration at line 1407 (on `main`) to `for stream, dst_multicast in zip(streams, dst_multicasts)`?

src/katsdpcontroller/product_controller.py
```diff
@@ -1625,6 +1625,11 @@ async def capture_start_stop(self, stream_name: str, *, start: bool) -> None:
                     product_config.GpucbfTiedArrayChannelisedVoltageStream,
                 ),
             ):
+                # NOTE: GpucbfTACV streams cannot be stopped if they feed a GpucbfTARV stream.
+                if _has_downstream_dependencies(stream, self.configuration.streams):
+                    raise FailReply(
+                        f"Cannot stop stream {stream_name!r} because it has downstream dependents"
+                    )
```
bmerry: Untested, but I think a cheaper test that looking for downstream dependencies is to
(a) move the `node = self.find_multicast_node(stream_name)` line that's further down above all this logic; then
(b) check `node.logical_node.initial_transmit_state` (if it's UP then we shouldn't stop it).

I'm also a little worried that using `_has_downstream_dependencies` like this will break on the case of SDP and CBF in a single subarray product because e.g. sdp.vis will be downstream of a baseline-correlation-products stream (generator.py correctly limits things to just tarv downstreams).

test/test_product_controller.py
```diff
@@ -2099,6 +2099,15 @@ async def test_capture_stop(self, client: aiokatcp.Client, sensor_proxy_client)
...
         # so the testing does not need to be as thorough.
         await self._configure_subarray(client, SUBARRAY_PRODUCT)
+
+        gpucbf_tarv_src_stream_name = "gpucbf_tied_array_channelised_voltage_0x_narrowband_vlbi"
+        with pytest.raises(
+            FailReply,
+            match=f"Cannot stop stream {gpucbf_tarv_src_stream_name!r} "
+            "because it has downstream dependents",
+        ):
+            await client.request("capture-stop", gpucbf_tarv_src_stream_name)
```
bmerry: There is a parametrised `test_request_fails` test where I think this could go.

src/katsdpcontroller/product_controller.py
```diff
@@ -1625,6 +1625,11 @@ async def capture_start_stop(self, stream_name: str, *, start: bool) -> None:
                     product_config.GpucbfTiedArrayChannelisedVoltageStream,
                 ),
             ):
+                # NOTE: GpucbfTACV streams cannot be stopped if they feed a GpucbfTARV stream.
```
bmerry: The ICD says that capture-start should fail as well (even though it would be a no-op). Lift this check code out of the `if start` block.

src/katsdpcontroller/generator.py
```diff
@@ -963,6 +963,22 @@ def _make_fgpu(
     return fgpu_group
 
 
+def _has_downstream_dependencies(
+    src_stream: product_config.Stream,
+    dst_streams: Sequence[product_config.Stream],
+    check_by_name: bool = False,
```
bmerry: Yuck. What is the motivation for having two different ways to do this test?

bmerry (changes requested):

src/katsdpcontroller/generator.py
amishatishpatel: I figured I should "support" either way because the `dst_streams` contents are explicitly of the same type as `src_stream`
- i.e. Someone could reasonably (want to) search by/for the object itself (?)

But also, re-testing it, for some reason when I initially searched by object itself it did not find the object as I expected. Looks like I can remove the `check_by_name`.

amishatishpatel: Ugh, I'm not entirely confident this helper is needed anymore. I trust you will flag accordingly.

src/katsdpcontroller/product_controller.py
```diff
@@ -1597,6 +1597,15 @@ async def capture_start_stop(self, stream_name: str, *, start: bool) -> None:
...
         nodes = list(self.find_nodes(streams=[stream]))
         args: List[Any] = []
+
+        node = self.find_multicast_node(stream_name)
+        if node.logical_node.initial_transmit_state == TransmitState.UP:
+            # NOTE: MeerKAT Extension CBF-CAM ICD outlines criteria for this. Additionally,
+            # GpucbfTACV streams cannot be stopped if they are required by GpucbfTARV streams.
+            raise FailReply(
+                f"Cannot stop stream {stream_name!r} because it is required by another stream"
+            )
```
amishatishpatel, line 1608: From your original comment - https://github.com/ska-sa/katsdpcontroller/pull/845#discussion_r3468365971
> Untested, but I think a cheaper test that looking for downstream dependencies is to
(a) move the `node = self.find_multicast_node(stream_name)` line that's further down above all this logic; then
(b) check `node.logical_node.initial_transmit_state` (if it's UP then we shouldn't stop it).
> I'm also a little worried that using `_has_downstream_dependencies` like this will break on the case of SDP and CBF in a single subarray product because e.g. sdp.vis will be downstream of a baseline-correlation-products stream (generator.py correctly limits things to just tarv downstreams).

I'm not 100% I've updated as you entirely envisioned. Again, trust you will shout either way.

test/test_product_controller.py
```diff
@@ -2099,6 +2101,7 @@ async def test_capture_stop(self, client: aiokatcp.Client, sensor_proxy_client)
         # Note: most of the code for capture-stop is shared with capture-start,
         # so the testing does not need to be as thorough.
         await self._configure_subarray(client, SUBARRAY_PRODUCT)
+
```
amishatishpatel: Sorry I'm not sure where this new line came from. Removing now.

src/katsdpcontroller/product_controller.py
```diff
@@ -1597,6 +1597,15 @@ async def capture_start_stop(self, stream_name: str, *, start: bool) -> None:
 
         nodes = list(self.find_nodes(streams=[stream]))
         args: List[Any] = []
+
+        node = self.find_multicast_node(stream_name)
+        if node.logical_node.initial_transmit_state == TransmitState.UP:
+            # NOTE: MeerKAT Extension CBF-CAM ICD outlines criteria for this. Additionally,
+            # GpucbfTACV streams cannot be stopped if they are required by GpucbfTARV streams.
+            raise FailReply(
+                f"Cannot stop stream {stream_name!r} because it is required by another stream"
```
bmerry: The logic looks good, but the message will be wrong for capture-start (should say "Cannot start stream...").

bmerry (changes requested):
Looks good - just needs to tweak the error message.

bmerry (approved):

PR #846, opened 2026-06-24
Improve get_command_argument

bmerry:
Make some of the code more readable by passing arguments to commands as:
```
[
    "--foo=bar",
    "--baz=spam",
    ...
]
```
rather than
```
[
    "--foo",
    "bar",
    "--baz",
    "spam",
]
```

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -810,31 +810,19 @@ def _make_fgpu(
             + taskset
             + [
                 "fgpu",
-                "--recv-interface",
-                "{interfaces[gpucbf].name}",
-                "--send-interface",
-                "{interfaces[gpucbf].name}",
-                "--send-packet-payload",
-                str(GPUCBF_PACKET_PAYLOAD_BYTES),
-                "--adc-sample-rate",
-                str(srcs[0].adc_sample_rate),
-                "--feng-id",
-                str(i),
-                "--array-size",
-                str(n_engines),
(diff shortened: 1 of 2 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1675,74 +1637,44 @@ def _make_vgpu(
...
     vgpu.capabilities.append("SYS_NICE")  # For schedrr
     if ibv:
         vgpu.capabilities.append("NET_RAW")
         vgpu.command += ["--recv-ibv"]
         if not configuration.options.develop.less_resources:
-            vgpu.command += [
-                "--recv-comp-vector",
-                "{cores[recv]}",
-            ]
+            vgpu.command += ["--recv-comp-vector={cores[recv]}"]
```
EatonEmmerich, line 1677: there are a bunch of entries here that seem to require the `f` prefix? it seems like the values weren't passed into a formatter before your change though.

EatonEmmerich (changes requested):

src/katsdpcontroller/generator.py
bmerry: They're not formatted at this point in the code, because we don't yet know which cores these will be. Rather, once the scheduler has allocated the task to specific cores on a specific host, the scheduler code will call `str.format` to fill in the information.

EatonEmmerich (approved):
LGTM

PR #847, opened 2026-06-29
Implement product-configure schema version 4.9

bmerry:
This should NOT be merged as is, as it doesn't implement version 4.8 and does not support the 4.8 features in 4.9 documents. That will come from #843.

See NGC-1779 and NGC-1992.

### DIFF
--- a/src/katsdpcontroller/product_config.py
+++ b/src/katsdpcontroller/product_config.py
@@ -69,6 +69,10 @@ BYTES_PER_WEIGHT = 1
 BYTES_PER_VFW = BYTES_PER_VIS + BYTES_PER_FLAG + BYTES_PER_WEIGHT
 
 
+def _is_power_two(x: int) -> bool:
+    return x > 0 and (x & (x - 1)) == 0
+
+
 def _url_n_endpoints(url: Union[str, yarl.URL]) -> int:
     """Return the number of endpoints in a ``spead://`` URL.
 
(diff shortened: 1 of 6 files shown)


src/katsdpcontroller/product_config.py
```diff
@@ -69,6 +69,10 @@
 BYTES_PER_VFW = BYTES_PER_VIS + BYTES_PER_FLAG + BYTES_PER_WEIGHT
 
 
+def _is_power_two(x: int) -> bool:
+    return x > 0 and (x & (x - 1)) == 0
```
EatonEmmerich, line 74: do we typically care about negative values?
I don't like that we have this and in a function called `_is_power_two`.
We already check the minimum for the config via yaml.
in which case would it actually be able to be negative?

bmerry: I guess the way it's used, it probably can't be negative, but I wanted to function to be correct independently of how it's used. If you remove the `x > 0` test then `_is_power_two(0)` returns true, which is wrong.

EatonEmmerich: In that sense, wouldn't `_is_power_two(1)` returning false be wrong?

bmerry: > In that sense, wouldn't `_is_power_two(1)` returning false be wrong?

If it did that would be wrong. But it doesn't, it returns true.

EatonEmmerich (commented):
LGTM, will await 4.8 before reviewing again (I see this MR is still in draft too)

test/test_product_config.py
```diff
@@ -371,6 +387,20 @@ def test_command_line_extra(self, config: Dict[str, Any]) -> None:
         dig = SimDigBasebandVoltageStream.from_config(Options(), "m000h", config, [], {})
         assert dig.command_line_extra == config["command_line_extra"]
 
+    def test_from_config_non_default(self, config: Dict[str, Any]) -> None:
+        """Test with default values being overridden."""
+        config["bits_per_sample"] = 6
+        config["samples_per_heap"] = 8192
+        dig = SimDigBasebandVoltageStream.from_config(Options(), "m000h", config, [], {})
```
EatonEmmerich, line 394: > `"m00h"`

Image

EatonEmmerich (approved):

PR #849, opened 2026-07-02
NGC-1555 - Eliminate device_attributes from Mesos GPU attributes- Final step

vincentseotlo:
Remove unused device_attribute


bmerry (commented):
This will need to wait until SDP have removed the attribute on all their deployed machines.

bmerry (approved):
LGTM, but don't merge yet until SDP have updated their machines.

PR #852, opened 2026-07-14
Hide per-engine sensors that are only needed for aggregation

bmerry:
This prevents them appearing in the device server's sensor list.

Most of the commits are refactoring of the SensorWatcher framework to replace one very complicated SensorWatcher with a simpler implementation which then gets used multiple times with different to achieve different things. I kept the history reasonably clean to aid piecemeal reviewing.

Closes NGC-2042

- [ ] Run qualification test to ensure they don't depend on any of these sensors.

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -950,7 +950,7 @@ def _make_fgpu(
 def _make_xbgpu(
     g: networkx.MultiDiGraph,
     configuration: Configuration,
-    sensors: SensorSet,
+    task_sensors: SensorSet,
     streams: Iterable[GpucbfXBStream],
 ) -> scheduler.LogicalNode:
     # Ensure that streams is a sequence, not just an iterable. Also put
@@ -976,8 +976,8 @@ def _make_xbgpu(
     ants = np.array(acv.input_labels).reshape(-1, 2)
     n_ants = ants.shape[0]
 
-    base_name = streams[0].name
-    xbgpu_group = LogicalGroup(f"xbgpu.{base_name}")
+    base_name = f"xb.{streams[0].name}"
+    xbgpu_group = LogicalGroup(base_name)
     g.add_node(xbgpu_group)
     dst_multicasts = []
     data_suspect_sensors = []
@@ -998,8 +998,8 @@ def _make_xbgpu(
             initial_status=Sensor.Status.NOMINAL,
         )
         data_suspect_sensors.append(data_suspect_sensor)
+        re_prefix = re.escape(f"{base_name}.") + "[0-9]+" + re.escape(f".{stream.name}.")
 
-        escaped_name = re.escape(stream.name)
         stream_sensors: List[Sensor] = [
             # The timestamps are simply ADC sample counts
             Sensor(
@@ -1125,20 +1125,20 @@ def _make_xbgpu(
                     initial_status=Sensor.Status.NOMINAL,
                 ),
                 SumSensor(
-                    sensors,
+                    task_sensors,
                     f"{stream.name}.xeng-clip-cnt",
                     "Number of visibilities that saturated",
-                    name_regex=re.compile(rf"{escaped_name}\.[0-9]+\.xeng-clip-cnt"),
+                    name_regex=re.compile(rf"{re_prefix}xeng-clip-cnt"),
                     n_children=stream.n_substreams,
                     auto_strategy=SensorSampler.Strategy.EVENT_RATE,
                     auto_strategy_parameters=(FAST_SENSOR_UPDATE_PERIOD, math.inf),
                 ),
                 SyncSensor(
-                    sensors,
+                    task_sensors,
                     f"{stream.name}.rx.synchronised",
                     "For the latest accumulation, was data present from all F-Engines "
                     "for all X-Engines",
-                    name_regex=re.compile(rf"{escaped_name}\.[0-9]+\.rx.synchronised"),
+                    name_regex=re.compile(rf"{re_prefix}rx.synchronised"),
                     n_children=stream.n_substreams,
                 ),
             ]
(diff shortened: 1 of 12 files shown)


test/test_sensor_proxy.py
```diff
@@ -128,7 +148,10 @@ async def client_and_watcher(
             renames={
                 "bytes-sensor": "custom-bytes-sensor",
                 "broadcast-sensor": ["copy01-broadcast-sensor", "copy02-broadcast-sensor"],
-            },
+            }
+            if not rewrite_readings
+            else None,
```
amishatishpatel, line 153: Interesting formatting.

test/test_sensor_proxy.py
```diff
@@ -43,6 +43,16 @@
...
 
 
+def _rewrite_readings(sensor: aiokatcp.Sensor, reading: aiokatcp.Reading):
+    if sensor.name == "prefix-int-sensor":
+        return aiokatcp.Reading(
+            timestamp=reading.timestamp + 0.5,
+            status=aiokatcp.Sensor.Status.WARN,
+            value=reading.value + 1,
+        )
+    return reading
```
amishatishpatel: Why did you put this up here? And e.g. not inside `TestSensorWatcher`

src/katsdpcontroller/sensor_proxy.py
```diff
@@ -177,15 +176,20 @@ def sensor_removed(self, name: str) -> None:
...
             rewritten_name = self.rewrite_name(name)[0]
             sensor = self.sensors[rewritten_name]
-            if rewritten_name.endswith(".gui-urls") and sensor.stype is bytes:
-                reading = self.rewrite_gui_urls(sensor, reading)
-        super().sensor_updated(name, reading.value, reading.status, reading.timestamp)
+            decoded_value = aiokatcp.decode(sensor.stype, value)
+            reading = aiokatcp.Reading(timestamp=timestamp, status=status, value=decoded_value)
+            reading = self.rewrite_readings(sensor, reading)
+            # TODO: move this functionality into aiokatcp so that we don't need
+            # to re-encode the value just for aiokatcp to decode it again.
```
amishatishpatel: Fair enough. Have you created a ticket for this?

src/katsdpcontroller/generator.py
```diff
@@ -1125,20 +1125,20 @@ def get_baseline_index(a1, a2):
                     initial_status=Sensor.Status.NOMINAL,
                 ),
                 SumSensor(
-                    sensors,
+                    task_sensors,
                     f"{stream.name}.xeng-clip-cnt",
                     "Number of visibilities that saturated",
-                    name_regex=re.compile(rf"{escaped_name}\.[0-9]+\.xeng-clip-cnt"),
+                    name_regex=re.compile(rf"{re_prefix}xeng-clip-cnt"),
```
amishatishpatel: Something feels weird about not seeing the dot/period hierarchical separator from stream to sensor.
- I guess it's understandable enough that it's part of the `prefix` (kinda like in katgpucbf's common recv/send)

src/katsdpcontroller/tasks.py
```diff
@@ -511,17 +511,30 @@ async def wait_ready(self):
...
+                        prefix,
+                        close_action=sensor_proxy.CloseAction.UNREACHABLE,
+                    ),
+                ]
+                for sensor_watcher in sensor_watchers:
+                    self.katcp_connection.add_sensor_watcher(sensor_watcher)
                 try:
-                    await sensor_watcher.synced.wait()
+                    for sensor_watcher in sensor_watchers:
+                        await sensor_watcher.synced.wait()
```
amishatishpatel, line 537: Any point in putting these in an `asyncio.gather` (when last did we use that)?

src/katsdpcontroller/generator.py
```diff
@@ -1454,22 +1454,23 @@ def get_baseline_index(a1, a2):
         ]:
             xbgpu.sensor_renames[name] = [f"{stream.name}.{i}.{name}" for stream in streams]
 
-        # Rename sensors that are relevant to the stream rather than the Pipeline
+        # Rename sensors that are relevant to the stream rather than the Pipeline,
+        # and hide sensors that are only used to create aggregates.
         for stream in streams:
             if isinstance(stream, product_config.GpucbfBaselineCorrelationProductsStream):
                 renames = ["chan-range", "rx.synchronised", "xeng-clip-cnt"]
+                hide = ["rx.synchronised", "xeng-clip-cnt"]
```
amishatishpatel, line 1462: Hang on, am I missing something here... `hide` contains two-thirds of `renames`?
- And you still refer to the per-engine `rx.synchronised` in `test_product_controller.py` 

I'll admit, it's been a journey refreshing my memory with this PR. I'll go through again in the morning to confirm if/how much I'm confused.

test/test_sensor_proxy.py
```diff
@@ -100,12 +109,10 @@ def __call__(self, sensor: aiokatcp.Sensor, reading: aiokatcp.Reading) -> None:
...
 @pytest.mark.timeout(5)
-class TestSensorProxyClient:
+class TestSensorWatcher:
     @pytest.fixture
-    def mirror(self, mocker) -> mock.MagicMock:
-        mirror = mocker.create_autospec(aiokatcp.DeviceServer, instance=True)
-        mirror.sensors = aiokatcp.SensorSet()
-        return mirror
+    def mirror(self) -> aiokatcp.SensorSet:
+        return aiokatcp.SensorSet()
```
amishatishpatel: I'll admit, I'm not a fan of this still being called `mirror`.
- I know you've added required/relevant type hints and updated usages, but
- I still see it (used) next to a DummyServer or Client and think it's mirroring that functionality.

I mean, it's "just another rename", but is it a bit ridiculous to call it e.g. `mirrored_sensors`?

amishatishpatel (commented):
A lot more to take in that I initially thought. I'm aiming to give it one more read-through tomorrow morning. 

In case you want to tackle this before you get on with the day, some minor comments below.

src/katsdpcontroller/sensor_proxy.py
```diff
@@ -185,16 +177,15 @@ def sensor_removed(self, name: str) -> None:
     def sensor_updated(
         self, name: str, value: bytes, status: aiokatcp.Sensor.Status, timestamp: float
     ) -> None:
-        super().sensor_updated(name, value, status, timestamp)
-        for rewritten_name in self.rewrite_name(name):
+        reading = aiokatcp.Reading(timestamp=timestamp, status=status, value=value)
+        if self.rewrite_gui_urls is not None:
+            # We don't allow custom renames, so this is guaranteed to
+            # have one exactly element.
+            rewritten_name = self.rewrite_name(name)[0]
```
amishatishpatel: Is it worth updating/trimming `rewrite_name` to cut out this step? The "putting into a list then only taking the first element" step.

In fact, looking at `rewrite_name` again is a bit weird... Do you know what scenario it would legitimately return `Sequence[str]`?

amishatishpatel: FWIW I don't see `rewrite_name` used anywhere else in katsdpcontroller. I also feel like this update is a pretty hard line to draw in how this works/is used. Not that I'm for/against it, or upset by it, just that if you're doing this it's worth considering reworking altogether?

test/test_sensor_proxy.py
bmerry: It could probably go there too, and you could ask the same about pretty much all the code in this file that isn't inside TestSensorWatcher or TestPrometheusWatcher e.g. DummyServer *could* be an inner class. But I reckon sub-consciously I was thinking that the TestSensorWatcher code contains the test itself (which does stuff to device servers and sensor watchers) while the code outside TestSensorWatcher is the code being tested (dummy implementations of device servers and sensor watcher callbacks).

src/katsdpcontroller/sensor_proxy.py
bmerry: Good call. I've filed NGC-2078.

src/katsdpcontroller/generator.py
bmerry: Oops, I forgot to remove them from `renames` when adding to `hide`. It won't make any practical difference because the set the `hide` rename attributes second and those will override those from `renames`, but I'll fix it anyway.

src/katsdpcontroller/sensor_proxy.py
bmerry: `self.renames.get` can return a sequence, and we use that when we have an engine-level sensor that we rewrite into a stream-level sensor for every stream handled by an engine. We now also use it to hide sensors by returning `[]`.

src/katsdpcontroller/tasks.py
amishatishpatel: I'll take this is a "No.".

amishatishpatel (approved):

bmerry (commented):
I've updated to use the new aiokatcp feature (released in 2.3.0).

amishatishpatel (approved):

PR #854, opened 2026-07-21
Pass ssl_option to _get_token

bmerry:
This is needed to make authentication with Harbor work when Harbor uses a local CA certificate that's in /etc/ssl/certs/ca-certificates.crt but not in the certifi trust store.

### DIFF
--- a/src/katsdpcontroller/scheduler.py
+++ b/src/katsdpcontroller/scheduler.py
@@ -1415,13 +1415,14 @@ class HTTPImageLookup(_RegistryImageLookup):
         realm: str,
         service: str,
         scope: str,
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/scheduler.py
EatonEmmerich, line 1412: shouldn't usages of get_token include updates of new ssl param?

bmerry: There is only one usage, and it's been updated (line 1475).

EatonEmmerich: my bad my quick read made it look like it was inside the updated function :ok_hand:

EatonEmmerich (approved):

PR #855, opened 2026-08-04
Fix vgpu multicast_in

bmerry:
It was specified with just the stream name, but the xb-engines
multicast_out has a separate entry for each substream, so they were not
matching. Also, xbgpu's multicast_out was accidentally using the stream object instead of the stream name.

Closes NGC-2091

### DIFF
--- a/src/katsdpcontroller/generator.py
+++ b/src/katsdpcontroller/generator.py
@@ -1632,7 +1632,9 @@ def _make_vgpu(
         scheduler.InterfaceRequest(
             "gpucbf",
             infiniband=ibv,
-            multicast_in={src_stream.name for src_stream in stream.src_streams},
+            multicast_in={
+                (src_stream.name, i) for src_stream in tacv for i in range(src_stream.n_substreams)
+            },
             multicast_out={stream.name},
         )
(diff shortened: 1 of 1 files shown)


src/katsdpcontroller/generator.py
```diff
@@ -1632,7 +1632,9 @@ def _make_vgpu(
         scheduler.InterfaceRequest(
             "gpucbf",
             infiniband=ibv,
-            multicast_in={src_stream.name for src_stream in stream.src_streams},
+            multicast_in={
+                (src_stream.name, i) for src_stream in tacv for i in range(src_stream.n_substreams)
```
EatonEmmerich, line 1636: are there some streams where the interface doesn't need to be defined on the controller level?

EatonEmmerich (approved):
Quick Q, LGTM otherwise.
I think this is a great candidate for a test, although I think that should come after the fix at some point rather.

src/katsdpcontroller/generator.py
bmerry: I'm afraid I don't understand the question.

PR #856, opened 2026-08-05
Remove bypass of data_timeout for V-engines

bmerry:
No longer needed after #845.


bmerry:
I'll still need to test this once Jenkins is back up to build it.

EatonEmmerich (approved):
approving so long, lgtm

PR #857, opened 2026-08-06
Add fake n-samples-per-frame

EatonEmmerich:


src/katsdpcontroller/fake_servers.py
```diff
@@ -520,6 +520,16 @@ def __init__(self, *args, **kwargs) -> None:
                 initial_status=Sensor.Status.NOMINAL,
             )
         )
+        self.sensors.add(
+            Sensor(
+                int,
+                "n-samples-per-frame",
+                "Number of samples per frame",
```
EatonEmmerich:
```suggestion
                "Number of samples per frame in each stream.",
```

bmerry (changes requested):
This shouldn't be in fake_servers.py. Add it alongside the other static sensors like pol-ordering and sideband-taps in generator._make_vgpu.

src/katsdpcontroller/fake_servers.py
```diff
@@ -520,6 +520,16 @@ def __init__(self, *args, **kwargs) -> None:
                 initial_status=Sensor.Status.NOMINAL,
             )
         )
+        self.sensors.add(
+            Sensor(
+                int,
+                "n-samples-per-frame",
+                "Number of samples per frame in the stream.",
+                units="samples/frame",
```
bmerry: Skip using units here. katcp units are normally used for physical units (time, speed, temperature etc).

src/katsdpcontroller/fake_servers.py
```diff
@@ -520,6 +520,16 @@ def __init__(self, *args, **kwargs) -> None:
...
+                default=2000,
```
bmerry: Use defaults.VGPU_SAMPLES_PER_FRAME

src/katsdpcontroller/generator.py
```diff
@@ -1617,8 +1619,26 @@ def _make_vgpu(
...
+            default=n_samples_per_frame,
+            initial_status=Sensor.Status.NOMINAL,
+        ),
     ]
 
+    for i, val in enumerate(threshold):
+        stream_sensors.append(
+            Sensor(
+                float,
+                f"quantisation-threshold-{i}",
```
bmerry: Missing the {stream.name} prefix.

src/katsdpcontroller/generator.py
```diff
@@ -1686,12 +1706,12 @@ def _make_vgpu(
...
             f"--send-bandwidth={stream.bandwidth}",
             f"--send-pols={','.join(stream.pols)}",
-            f"--send-samples-per-frame={defaults.VGPU_SAMPLES_PER_FRAME}",
+            f"--send-samples-per-frame={n_samples_per_frame}",
             f"--send-station={stream.station_id}",
             f"--fir-taps={defaults.VGPU_FIR_TAPS}",
             f"--hilbert-taps={defaults.VGPU_HILBERT_TAPS}",
             f"--passband={defaults.VGPU_PASSBAND}",
-            f"--threshold={defaults.VGPU_THRESHOLD}",
+            f"--threshold={threshold[0]}",  # TODO: Change VGPU threshold arg to be a list
```
bmerry: I don't see any need for that TODO, since it is hard-coded to only support 2-bit output anyway. If you wanted to future-proof this code, you could something like

```suggestion
            f"--threshold={','.join(str(t) for t in threshold)}",
```
(untested, not sure if such constructs are allowed inside an f-string)

src/katsdpcontroller/generator.py
```diff
@@ -1617,8 +1619,26 @@ def _make_vgpu(
...
+        Sensor(
+            int,
+            f"{stream.name}.n-samples-per-frame",
+            "Number of samples per frame in the stream.",
+            default=n_samples_per_frame,
+            initial_status=Sensor.Status.NOMINAL,
+        ),
     ]
 
+    for i, val in enumerate(threshold):
```
bmerry:
```suggestion
    for i, val in enumerate(threshold, start=1):
```
See the ICD: the range for i starts at 1.

bmerry (changes requested):

src/katsdpcontroller/fake_servers.py
EatonEmmerich: I'm not 100% following the statement (example is quantisation threshold a physical unit?)
would bandwidth be physical?
removed both samples-per-frame and quantisation threshold

src/katsdpcontroller/generator.py
EatonEmmerich: I'm more concerned that the comma seperated value won't be interpreted correctly in the v engine though, i'd rather just remove it than leave a footgun

src/katsdpcontroller/fake_servers.py
bmerry: Bandwidth is physical (Hz). The quantisation threshold is unitless (it's specified relative to the standard deviation). It's entirely a matter of convention / historical consistency, but generally if the value is not dependent on SI units we tend not to use the `units` field.

src/katsdpcontroller/generator.py
```diff
@@ -1617,8 +1619,26 @@ def _make_vgpu(
...
+            initial_status=Sensor.Status.NOMINAL,
+        ),
     ]
 
+    for i, val in enumerate(threshold, start=1):
+        stream_sensors.append(
+            Sensor(
+                float,
+                f"{stream.name}.quantisation-threshold-{i}",
+                "Quantisation threshold per value for all streams.",
```
bmerry: Let's match it with the ICD description
```suggestion
                "Voltage threshold between the quantisation levels, in units of RMS
voltage.",
```

bmerry (changes requested):

